PR #5, opened 2016-03-08
inherit FormParam and QueryParam from Property

czardoz:


czardoz:
GitHub gave me a nice button to click, and then added the merge commit :@

PR #6, opened 2016-03-08
Implementation of a generic toJSON

czardoz:
This is a WIP, do not merge it yet.

### DIFF
--- a/lib/collection/collection.js
+++ b/lib/collection/collection.js
@@ -97,6 +97,21 @@ _.inherit((
         });
     }), ItemGroup);
 
+_.extend(Collection.prototype, {
+    toJSON: function () {
+        return {
+            info: {
+                id: this.id,
+                name: this.name,
+                version: this.version
+            },
+            event: this.events.toJSON(),
+            variable: this.variables.toJSON(),
+            item: this.items.toJSON()
+        };
+    }
+});
+
 module.exports = {
     Collection: Collection
 };
--- a/lib/collection/collection.js
+++ b/lib/collection/collection.js
@@ -108,8 +108,8 @@ _.extend(Collection.prototype, {
                 version: this.version,
                 schema: SCHEMA_URL
             },
-            event: this.events.toJSON(),
-            variable: this.variables.toJSON(),
+            event: this.events ? this.events.toJSON() : undefined,
+            variable: this.variables ? this.variables.toJSON() : undefined,
             item: this.items.toJSON()
         };
     }
--- a/lib/collection/collection.js
+++ b/lib/collection/collection.js
@@ -2,7 +2,9 @@ var _ = require('../util').lodash,
     ItemGroup = require('./item-group').ItemGroup,
     VariableList = require('./variable-list').VariableList,
 
-    Collection; // constructor
+    Collection, // constructor
+
+    SCHEMA_URL = 'https://schema.getpostman.com/json/collection/v2.0.0/collection.json';
 
 _.inherit((
     /**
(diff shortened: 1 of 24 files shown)


czardoz:
@shamasis ready for your review

lib/collection/property-list.js
```diff
@@ -272,6 +272,42 @@ _.extend(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
         }
+    },
+
+    toJSON: function () {
+        if (!this.count()) {
+            return;
+        }
+
+        return _.map(this.members, function (member) {
+            return _.reduce(member, function (accumulator, value, key) {
```
czardoz: This logic is common to the PropertyList and PropertyBase. Putting it in `util.js` is not an option because of cyclic dependency issues.. will figure out a better place for this. 

Suggestions welcome :)

shamasis: It's okay here as of now.

lib/collection/collection.js
```diff
@@ -97,6 +99,22 @@ _.inherit((
...
     }), ItemGroup);
 
+_.extend(Collection.prototype, {
+    toJSON: function () {
+        return {
+            info: {
+                id: this.id,
+                name: this.name,
+                version: this.version,
+                schema: SCHEMA_URL
```
shamasis: Another indication that schema should reside in this project itself.

lib/collection/property-base.js
```diff
@@ -1,5 +1,6 @@
 var _ = require('../util').lodash,
     Description = require('./description').Description,
+    PropertyList = require('./property-list').PropertyList,
```
shamasis: Base cannot have list.

lib/collection/property-base.js
```diff
@@ -35,6 +36,38 @@ PropertyBase = function PropertyBase (options) {
...
+            }
+
+            // Handle Strings
+            if (_.isString(value)) {
+                accumulator[key] = value;
+                return accumulator;
+            }
+
+            // Everything else
+            accumulator[key] = _.clone(value);
```
shamasis: Clone is shallow. Should it not be cloneDeep? Think objects. But problem with clone deep is that it will also start cloning properties encountered as grandchild. You'll need to implement your own clone deep that respects toJSON.

lib/collection/property-base.js
czardoz: Base needs to check if an object is a PropertyList

czardoz: Actually it is only needed because we internally have plural names for lists, but the collection format has singular names. [headers vs header, items vs item, etc]

postman-admin: Use flags in prototype?
On Thu, 10 Mar 2016 at 13:40, Aniket Panse <notifications@github.com> wrote:

> In lib/collection/property-base.js
> <https://github.com/postmanlabs/postman-collection/pull/6#discussion_r55645245>
> :
>
> > @@ -1,5 +1,6 @@
> >  var _ = require('../util').lodash,
> >      Description = require('./description').Description,
> > +    PropertyList = require('./property-list').PropertyList,
>
> Actually it is only needed because we internally have plural names for
> lists, but the collection format has singular names. [headers vs header,
> items vs item, etc]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/postmanlabs/postman-collection/pull/6/files#r55645245>
> .
>
-- 
--
Shamasis Bhattacharya

VPE / Postman

shamasis:
checking

PR #9, opened 2016-03-15
Private environment access API

shamasis:
`VariableList.prototype.env` added;

```
    /**
     * This function allows to manipulate the environment chain in this variable list.
     * @private
     *
     * @param {Number} [index] If an index value is provided, this function returns the JSON representation of the
     * particular environment. In its absence, all environments are returned within an array.
     * @param {(Object|(Array<Variable>)|Null)=} [assign] A list of variables can be passed on for replacement in the
     * environment. If a replacement is sent, then the function does not return any JSON.
     * @param {Boolean=} [clear=false] If set to true, then the environment is cleared before new set of assignments are
     * done. The assignment can be sent a `null` if environment is needed to be cleared alone.
     *
     * @returns {Object|undefined}
     *
     * @example
     * myCollection.variables.env(); // gets all environments as array
     * myCollection.variables.env(1); // gets the second environment as Object
     * myCollection.variables.env(0, null, true); // cleares the first environment (does not delete the layer though)
     * myCollection.variables.env(0, {
     *     var1: "variable-value"
     * }); // assigns `var1` variable to the first environment layer
     * myCollection.variables.env(1, {
     *     var2: "variable-value-2"
     * }); // clears the 2nd layer and then assigns `var2` variable to it
     * myCollection.variables.env(0, [
     *     new Variable({ id: "var3", value: "value-3"}),
     *     new Variable({ id: "var4", value: "value-4"})
     * ]); // accepts array of variables property ass assignment
     */
```

### DIFF
--- a/lib/collection/variable-list.js
+++ b/lib/collection/variable-list.js
@@ -1,6 +1,17 @@
 var _ = require('../util').lodash,
     PropertyList = require('./property-list').PropertyList,
+    PropertyBase = require('./property-base').PropertyBase,
     Variable = require('./variable').Variable,
+
+    /**
+     * Convert object to JSON using the base converter
+     * @private
+     * @param {Object} obj
+     * @returns {Object}
+     */
+    toJSON = function (obj) {
+        return PropertyBase.prototype.toJSON.call(obj);
+    },
 
     /**
      * Maintain a list of types that are native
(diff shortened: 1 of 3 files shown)


lib/collection/variable-list.js
```diff
@@ -74,6 +85,58 @@ _.extend(VariableList.prototype, /** @lends VariableList.prototype */ {
...
+     * }); // clears the 2nd layer and then assigns `var2` variable to it
+     * myCollection.variables.env(0, [
+     *     new Variable({ id: "var3", value: "value-3"}),
+     *     new Variable({ id: "var4", value: "value-4"})
+     * ]); // accepts array of variables property ass assignment
+     *
+     * @todo
+     * - convert this to an environment instance (coded within this module) and then add an easy API on it
+     */
+    env: function (index, assign, clear) {
```
czardoz: This function does a lot of things.. we should later add wrappers that internally call this function.

PR #11, opened 2016-03-17
Url Parser Fixes

shamasis:
This PR simplifies the code of URL parser and also fixes a few bugs as mentioned in the commits. It also improves the handling of path segments.

It also adds a bunch of unit tests on the `Url.parse` function that are primarily in and around IP address type URLs


czardoz:
:cake:

PR #12, opened 2016-03-20
Added draft and private `Property#toObjectResolved` to procure an Obj…

shamasis:
Added draft and private `Property#toObjectResolved` to procure an Object representation of a property with its variables resolved.

Calling `.toObjectResolved()` on any Property causes it to traverse the parent tree and procure the list of variables that it would substitute with.

This is a draft + private function

### DIFF
--- a/lib/collection/property.js
+++ b/lib/collection/property.js
@@ -96,6 +96,35 @@ _.extend(Property.prototype, /** @lends Property.prototype */ {
     describe: function (content, type) {
         (Description.isDescription(this.description) ? this.description : (this.description = new Description()))
             .update(content, type);
+    },
+
+    /**
+     * Returns an object representation of the Property with its variable references substituted.
+     * @private
+     * @draft
+     *
+     * @param {VariableList=} [variables] - One can specifically provide a VariableList for doing the substitution. In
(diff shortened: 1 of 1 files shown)


lib/collection/property.js
```diff
@@ -96,6 +96,35 @@ _.extend(Property.prototype, /** @lends Property.prototype */ {
...
+        // We see if variables can be found in this object itself.
+        !variables && (variables = variableSourceObj.variables);
+
+        // We then check if variables can be procured from the parent tree
+        while ((variableSourceObj = variableSourceObj.__parent)) {
+            variables = variableSourceObj.variables;
+        }
+
+        if (!variables) { // worst case = no variable param and none detected in tree or object
+            throw Error('Unable to resolve variables. Require a List type property for variable auto resolution.');
```
czardoz: There are cases where collections may not contain variables.. Can we assume an empty set by default?

shamasis: It is an empty set by default.

The error occurs when you try to resolve an orphan object or an object that has no discernible parent. E.g. item.request - request has no .__parent. As such for them, the variable object needs to be passed as a reference. Else it throws error.

PR #21, opened 2016-03-25
Added ability of Property.toObjectResolved to accept additional override sources

shamasis:
This PR adds the functionality to provide additional set of objects to be used for variable resolution in `Property#toObjectResolved`. This can hypothetically allow the use case where environment and global variables are NOT sent to the SDK and instead managed externally and sent to the resolution function.

Usage: 
```
myPropertyInstance.toObjectResolved(null, [{var1: "value1"}]);
```

Out of context: This PR also changes the internal flag `_postman_isGroup` to `_postman_propertyIsList` for keeping them in line with `_postman_propertyXYZ` flags

### DIFF
--- a/lib/collection/property.js
+++ b/lib/collection/property.js
@@ -103,13 +103,14 @@ _.extend(Property.prototype, /** @lends Property.prototype */ {
      * @private
      * @draft
      *
-     * @param {VariableList=} [variables] - One can specifically provide a VariableList for doing the substitution. In
+     * @param {?VariableList=} [variables] - One can specifically provide a VariableList for doing the substitution. In
      * the event one is not provided, the variables are taken from this object or one from the parent tree.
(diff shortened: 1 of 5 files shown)


lib/collection/property.js
```diff
@@ -103,13 +103,14 @@ _.extend(Property.prototype, /** @lends Property.prototype */ {
      * @private
      * @draft
      *
-     * @param {VariableList=} [variables] - One can specifically provide a VariableList for doing the substitution. In
+     * @param {?VariableList=} [variables] - One can specifically provide a VariableList for doing the substitution. In
```
shamasis: The `?` before variable type jsdoc indicates that it is a `nullable` type.

czardoz: :+1:

PR #22, opened 2016-04-04
added http reason phrases

czardoz:


lib/collection/response.js
```diff
@@ -36,7 +112,7 @@ _.extend(Response.prototype, /** @lends Response.prototype */ {
             /**
              * @type {String}
              */
-            status: options.status,
+            status: options.status ? options.status : HTTP_REASON_PHRASES[options.code],
```
shamasis: what happens if `options.status === 0`

lib/collection/response.js
```diff
@@ -5,7 +5,83 @@ var _ = require('../util').lodash,
...
+    Response,
+
+    /**
+     * Enum for all the HTTP Reason phrases
+     *
+     * @readonly
+     * @enum {string}
+     * @memberOf Response
+     */
+    HTTP_REASON_PHRASES = {
```
shamasis: We can actually move this to a static variable inside `Response`

PR #27, opened 2016-05-10
Added ability to substitute function variables, and tests

czardoz:


lib/collection/property.js
```diff
@@ -11,7 +11,8 @@ var _ = require('../util').lodash,
     nativeTypes = {
         'string': true,
         'number': true,
-        'boolean': true
+        'boolean': true,
+        'function': true
```
shamasis: remove this.

lib/collection/variable-list.js
```diff
@@ -38,7 +39,17 @@ _.inherit((
             var proxies = this._environments;
             environments.reduce(function (parent, child) {
                 return (proxies.push(VariableList.proxy(parent, VariableList.objectify(child))), _.last(proxies));
-            }, {}); // @todo: this blank object can have global default
+            }, {
+                $guid: function () {
+                    return uuid.v4();
+                },
+                $timestamp: function () {
+                    return Math.round(new Date().getTime() / 1000);
```
shamasis: use Date.now()

lib/collection/variable-list.js
```diff
@@ -38,7 +39,17 @@ _.inherit((
             var proxies = this._environments;
             environments.reduce(function (parent, child) {
                 return (proxies.push(VariableList.proxy(parent, VariableList.objectify(child))), _.last(proxies));
-            }, {}); // @todo: this blank object can have global default
+            }, {
```
shamasis: add a todo that these needs to be able to have external definitions. else it is super hard to find where they are defined.

PR #29, opened 2016-05-20
Added recursive variable resolution capabilities

kunagpal:
* Delimited iterative resolution depth to 20 cycles
* Added tests for recursion depth and cyclic resolution handling

@shamasis @czardoz

### DIFF
--- a/lib/collection/property.js
+++ b/lib/collection/property.js
@@ -170,9 +170,11 @@ _.extend(Property, /** @lends Property */ {
      * @returns {String}
      */
     replaceSubstitutions: function (str, variables) {
-        var loopCounter = -1,
-            resolvedStr = '',
+        var resolvedStr,
+            replaced = true,
+            loopCounter = 0,
             onReplace = function (match, token) {
+                replaced = true;
                 var r = _.findValue(variables, token);
                 r && _.isFunction(r) && (r = r());
                 r && _.isFunction(r.toString) && (r = r.toString());
@@ -187,7 +189,8 @@ _.extend(Property, /** @lends Property */ {
         // extract all arguments to get the list of variables
         !_.isArray(variables) && (variables = _.tail(arguments));
 
-        while (resolvedStr != str && ++loopCounter < 20) {
+        while (replaced && ++loopCounter < 21) {
+            replaced = false;
             resolvedStr = str;
             str = str.replace(this.REGEX_EXTRACT_VARS, onReplace);
         }
--- a/lib/collection/property.js
+++ b/lib/collection/property.js
@@ -162,6 +162,14 @@ _.extend(Property, /** @lends Property */ {
     REGEX_EXTRACT_VARS: /\{\{([^}]*?)}}/g,
 
     /**
+     * Upper iteration limit for recursive variable resolution. The resolver runs till a count of one less than this.
+     * @private
+     * @readOnly
+     * @type {Integer}
+     */
+    MAX_RESOLUTION_LIMIT: 21,
+
+    /**
      * This function accepts a string followed by a number of variable sources as arguments. One or more variable
      * sources can be provided and it will use the one that has the value in left-to-right order.
      *
@@ -169,18 +177,8 @@ _.extend(Property, /** @lends Property */ {
      * @param {VariableList|Object|Array.<VariableList|Object>} variables
      * @returns {String}
      */
+    // @todo: improve algorithm via variable replacement caching
     replaceSubstitutions: function (str, variables) {
-        var resolvedStr,
-            replaced = true,
-            loopCounter = 0,
-            onReplace = function (match, token) {
-                replaced = true;
-                var r = _.findValue(variables, token);
-                r && _.isFunction(r) && (r = r());
-                r && _.isFunction(r.toString) && (r = r.toString());
-                return nativeTypes[(typeof r)] ? r : match;
-            };
-
         // if there is nothing to replace, we move on
         if (!(str && _.isString(str))) {
             return str;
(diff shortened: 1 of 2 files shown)


lib/collection/property.js
```diff
@@ -170,6 +170,15 @@ _.extend(Property, /** @lends Property */ {
      * @returns {String}
      */
     replaceSubstitutions: function (str, variables) {
+        var loopCounter = -1,
```
czardoz: can you fix the loop iteration, so that it starts at zero?

lib/collection/property.js
```diff
@@ -178,12 +187,12 @@ _.extend(Property, /** @lends Property */ {
         // extract all arguments to get the list of variables
         !_.isArray(variables) && (variables = _.tail(arguments));
 
-        return str.replace(this.REGEX_EXTRACT_VARS, function (match, token) {
-            var r = _.findValue(variables, token);
-            r && _.isFunction(r) && (r = r());
-            r && _.isFunction(r.toString) && (r = r.toString());
-            return nativeTypes[(typeof r)] ? r : match;
-        });
+        while (resolvedStr != str && ++loopCounter < 20) {
```
czardoz: The original function was written with recursion in mind. Any particular reason you're using iterations?

czardoz: Also, can we avoid comparing these strings? Some people store response bodies as environment variables, so they can be pretty huge.

lib/collection/property.js
kunagpal: Sure thing, will do.

lib/collection/property.js
kunagpal: 1. Given that a fixed limit of 20 levels of depth checking has to be performed, a loop would be a better choice. This can be shifted to recursion with a fair amount of TCO, but a loop is more readable here.

2. I'm not quite sure how else the check for any remaining possible substitutions would work.

kunagpal:
@shamasis @czardoz A general note: The current build is failing due to issues with the environment variable resolution of [doc publishing keys](https://github.com/postmanlabs/postman-collection/blob/develop/.travis.yml#L6-7)

lib/collection/property.js
```diff
@@ -170,6 +170,15 @@ _.extend(Property, /** @lends Property */ {
      * @returns {String}
      */
     replaceSubstitutions: function (str, variables) {
+        var loopCounter = -1,
+            resolvedStr = '',
```
czardoz: You don't need to initialize it to a blank string, just declaring the var should be enough

shamasis:
@kunagpal tests seemed to have failed.

lib/collection/property.js
```diff
@@ -170,6 +170,17 @@ _.extend(Property, /** @lends Property */ {
      * @returns {String}
      */
     replaceSubstitutions: function (str, variables) {
+        var resolvedStr,
+            replaced = true,
+            loopCounter = 0,
+            onReplace = function (match, token) {
```
shamasis: this function is the core and is closed in a closure and as such it is no longer testable.

lib/collection/property.js
```diff
@@ -178,12 +189,13 @@ _.extend(Property, /** @lends Property */ {
         // extract all arguments to get the list of variables
         !_.isArray(variables) && (variables = _.tail(arguments));
 
-        return str.replace(this.REGEX_EXTRACT_VARS, function (match, token) {
-            var r = _.findValue(variables, token);
-            r && _.isFunction(r) && (r = r());
-            r && _.isFunction(r.toString) && (r = r.toString());
-            return nativeTypes[(typeof r)] ? r : match;
-        });
+        while (replaced && ++loopCounter < 21) {
```
shamasis: also, this `21` value is an important thing in this feature's context and as such difficult to configure and find. it should move out as a static variable in Property.

lib/collection/property.js
```diff
@@ -170,6 +170,17 @@ _.extend(Property, /** @lends Property */ {
      * @returns {String}
      */
     replaceSubstitutions: function (str, variables) {
```
shamasis: also, please add a `// @todo: improve algorithm by doing variable level replacement cacheing` since the current algorithm is a dfs and can receive some improvements later.

lib/collection/property.js
```diff
@@ -170,6 +170,17 @@ _.extend(Property, /** @lends Property */ {
...
+        var resolvedStr,
+            replaced = true,
+            loopCounter = 0,
+            onReplace = function (match, token) {
+                replaced = true;
+                var r = _.findValue(variables, token);
+                r && _.isFunction(r) && (r = r());
+                r && _.isFunction(r.toString) && (r = r.toString());
```
shamasis: The above two lines can join as (reducing one boolean check)

```
r && _.isFunction(r) && (r = r()) && _.isFunction(r.toString) && (r = r.toString());
```

lib/collection/property.js
shamasis: The definition of replace function should go down after current line 190 since the nature of `variables` is modified at that line and also there is a validation function bailout at line 185 and there is no point doing a function definition if the validations fail.

lib/collection/property.js
```diff
@@ -170,6 +170,17 @@ _.extend(Property, /** @lends Property */ {
      * @returns {String}
      */
     replaceSubstitutions: function (str, variables) {
+        var resolvedStr,
+            replaced = true,
+            loopCounter = 0,
```
shamasis: - loop assignment 0 is here while definition is down below.

lib/collection/property.js
shamasis: suggestion -  we could do a `> 0` reverse counter.

czardoz:
Ignoring the build failure, and merging this. The failure is because the PR is created from a fork.

PR #31, opened 2016-05-23
fixed the handling of function variables in a URL

czardoz:


lib/collection/url.js
```diff
@@ -307,7 +307,7 @@ _.extend(Url, /** @lends Url */ {
         }
 
         // extract the query string
-        p.query = _.get(url.match(/^\?([^#$]+)/), '[1]');
+        p.query = _.get(url.match(/^\?([^#]+)/), '[1]');
```
czardoz: What was the `$` for? EOL?

czardoz:
Do not merge yet, investigating why this is needed in the first place

lib/collection/url.js
shamasis: Yes. But seems not so correct now that I see it. There is a + operator. Can we add the failing test case and then the passing one?

czardoz:
Fixed everything up, with tests. There's no failing test, @shamasis because the failure cannot be reproduced anymore.

PR #40, opened 2016-06-05
Added ability to populate from an object

shamasis:
- Update PropertyList#add function to be a bit more intelligent and use the source Type.create function to add a new item whenever one is available.
- Updated PropertyList#populate function to accept an object as well and treat individual keys in them as separate items to be sent to PropertyList.add(). This allows simple object to be used as propertylist source for properties that support Property.create
- Added Header.create and updated the constructor to accept value, key as construction parameter

All these together allow headers to be now sent as object to `.populate` of PropertyList of headers

Caveat: tests pass, but not extensively tested whether it affects populates of other types. less likely though since `Property.create` is checked before treating it differently

### DIFF
--- a/lib/collection/property-list.js
+++ b/lib/collection/property-list.js
@@ -118,7 +118,9 @@ _.extend(PropertyList.prototype, /** @lends PropertyList.prototype */ {
     add: function (item) {
         if (!item) { return; } // do not proceed on empty param
         // create new instance of the item based on the type specified if it is not already
-        this.insert((item.constructor === this.Type) ? item : new this.Type(item));
+        this.insert((item.constructor === this.Type) ? item :
+            // if the prperty has acreate static function, use it.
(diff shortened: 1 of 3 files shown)


lib/collection/property-list.js
```diff
@@ -164,7 +166,10 @@ _.extend(PropertyList.prototype, /** @lends PropertyList.prototype */ {
         // if Type supports parsing of string headers then do it before adding it.
         _.isString(items) && _.isFunction(this.Type.parse) && (items = this.Type.parse(items));
         // add a single item or an array of items.
-        _.each(_.isArray(items) ? items : [items], this.add, this);
+        _.each(_.isArray(items) ? items :
+            // if population is not an array, we send this as single item in an array or send each property separately
```
czardoz: Do we really need to be so magical about adding things to a list?

PR #54, opened 2016-06-17
added a property to determine whether headers were added as a byprodu…

czardoz:
Added a property to determine whether headers were added as a byproduct of the sdk functions


shamasis:
Question is, how is this unset once an SDK consumer updates or edits a header?

czardoz:
There's no use-case where they might want to update/edit a header added by the system. The Auth helpers  _remove_ headers before adding them. 

If someone wants to have the freedom to update them though, they can either replace the header or just change the `system` flag. DO you feel the need to have a sugar function for this?

PR #59, opened 2016-06-28
ensure that btoa is called directly, otherwise Chrome throws an illeg…

czardoz:
…al invocation error


czardoz:
Merging because @shamasis wrote this in the first place :smile:

PR #62, opened 2016-07-20
added a check to see if request body is empty

czardoz:


lib/collection/request-body.js
```diff
@@ -120,6 +126,32 @@ _.extend(RequestBody.prototype, /** @lends RequestBody.prototype */ {
...
+
+    /**
+     * If the request body is set to a mode, but does not contain data, then we should not be sending it.
+     *
+     * @returns {Boolean}
+     */
+    isEmpty: function () {
+        var mode = this.mode,
+            data;
+        // If there's no Mode set, treat it as an empty body.
```
shamasis: whitespace before this

lib/collection/request-body.js
```diff
@@ -120,6 +126,32 @@ _.extend(RequestBody.prototype, /** @lends RequestBody.prototype */ {
...
             return (this.raw && _.isFunction(this.raw.toString)) ? this.raw.toString() : EMPTY;
         }
+    },
+
+    /**
+     * If the request body is set to a mode, but does not contain data, then we should not be sending it.
+     *
+     * @returns {Boolean}
+     */
+    isEmpty: function () {
```
shamasis: just a thought

```
var mode = this.mode,
    data = mode && this[mode];

return !data || !(_.isString(data) ? data.length : (_.isFunction(data.count) && data.count()));
```

PR #65, opened 2016-07-22
Added support for response size

vegetableman:
Initial commit. 

This checks for content-length header, if present, adds the size to the header size. Else, it adds the content size. 

@czardoz  @shamasis 

Tests are pending. Will add them after we are good to go with the changes.

### DIFF
--- a/lib/collection/header.js
+++ b/lib/collection/header.js
@@ -1,7 +1,7 @@
-var _ = require('../util').lodash,
+var util = require('../util'),
+    _  = util.lodash,
     PropertyBase = require('./property-base').PropertyBase,
     PropertyList = require('./property-list').PropertyList,
-
     Header;
 /**
  * @typedef Header~definition
@@ -196,6 +196,41 @@ _.extend(Header, /** @lends Header */ {
         var args = Array.prototype.slice.call(arguments);
         args.unshift(Header);
         return new (Header.bind.apply(Header, args))();
+    },
+
+    /**
+     * Gets value from an Array or {@link PropertyList} of Headers by filtering through key.
+     *
+     * @param {Array|PropertyList<Header>} headers
+     * @param {string} key
+     * @returns {string}
+     */
+    headerValue: function (headers, key) {
+        if (!_.isArray(headers) && !PropertyList.isPropertyList(headers)) {
+            return undefined;
+        }
+        var header = headers.find(function (header) {
+            return header.key === key;
+        });
+        return header ? header.value : undefined;
+    },
+
+    /**
+     * Gets size of an Array or {@link PropertyList} of Headers.
+     *
+     * @param {Array|PropertyList<Header>} headers
+     * @returns {Number}
+     */
+    size: function (headers, statusCode, reason) {
+        if (!_.isArray(headers) && !PropertyList.isPropertyList(headers)) {
+            return 0;
+        }
+        // https://tools.ietf.org/html/rfc7230#section-3.1.2
+        // status-line = HTTP-version SP status-code SP reason-phrase CRLF
+        var raw = 'HTTP/X.X' + ' ' + statusCode + ' ' + reason + '\r\n';
+        raw += this.unparse(headers, '\r\n');
+        raw += '\r\n\r\n';
+        return raw.length;
     }
 });
 
--- a/lib/collection/response.js
+++ b/lib/collection/response.js
@@ -211,6 +211,30 @@ _.extend(Response.prototype, /** @lends Response.prototype */ {
         // we create the body string first from stream and then fallback to body
         return 'data:' + mime.type + ';base64, ' + (((this.stream != null) &&
             util.bufferOrArrayBufferToBase64(this.stream)) || ((this.body != null) && util.btoa(this.body)) || E);
+    },
+
+    /**
+     * Get the response size
+     * @return {Number}
+     * @todo write unit tests
+     */
+    size: function () {
+        var sizeInfo = {},
+            contentEncoding = Header.headerValue(this.headers, 'Content-Encoding'),
+            contentLength = Header.headerValue(this.headers, 'Content-Length');
+
+        // if 'Content-Length' header is present and encoding is of type gzip/deflate
+        if (contentLength && /^(gzip|deflate)$/.test(contentEncoding) && util.isNumeric(contentLength)) {
+            sizeInfo.body = _.parseInt(contentLength, 10);
+        }
+        // else size of body is added
+        else {
+            sizeInfo.body = this.stream ? this.stream.byteLength : (this.body && this.body.length || 0);
+        }
+
+        // size of header is added
+        sizeInfo.header = Header.size(this.headers, this.code, this.reason());
+        return sizeInfo;
     }
 });
 
--- a/lib/util.js
+++ b/lib/util.js
@@ -233,6 +233,15 @@ util = {
         }
 
         return base64;
+    },
+
+    /**
+     * Check whether a value is number-like
+     * https://github.com/lodash/lodash/issues/1148#issuecomment-141139153
+     * @return {Boolean}
+     */
+    isNumeric: function (n) {
+        return !isNaN(parseFloat(n)) && isFinite(n);
     }
 };
 
--- a/test/fixtures/index.js
+++ b/test/fixtures/index.js
@@ -255,6 +255,16 @@ module.exports = {
                 value: 'somevalue'
             }
         ]
+    },
+    responseData1: {
+        code: 200,
+        body: '{"status":200}',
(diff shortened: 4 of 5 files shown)


lib/collection/header.js
```diff
@@ -190,6 +192,38 @@ _.extend(Header, /** @lends Header */ {
...
+    /**
+     * Gets value from an Array or {@link PropertyList} of Headers by filtering through key.
+     *
+     * @param {Array|PropertyList<Header>} headers
+     * @param {string} key
+     * @returns {string}
+     */
+    getHeaderValue: function(headers, key) {
+        if (!_.isArray(headers) && !PropertyList.isPropertyList(headers)) {
+            throw new Error('It is only possible to unparse PropertyList or Array of headers.');
```
czardoz: No throws in the SDK

lib/collection/header.js
```diff
@@ -190,6 +192,38 @@ _.extend(Header, /** @lends Header */ {
...
+        }
+        var header = _.find(headers, function (header) {
```
czardoz: In case you have a PropertyList, you should use http://www.postmanlabs.com/postman-collection/PropertyList.html#find

lib/collection/header.js
```diff
@@ -190,6 +192,38 @@ _.extend(Header, /** @lends Header */ {
...
+            return header.key === key;
+        });
+        return header ? header.value : null;
```
czardoz: `undefined` instead of null

lib/collection/header.js
```diff
@@ -190,6 +192,38 @@ _.extend(Header, /** @lends Header */ {
...
+
+    /**
+     * Gets size of an Array or {@link PropertyList} of Headers.
+     *
+     * @param {Array|PropertyList<Header>} headers
+     * @returns {Number}
+     */
+    getSize: function(headers) {
+        if (!_.isArray(headers) && !PropertyList.isPropertyList(headers)) {
+            throw new Error('It is only possible to unparse PropertyList or Array of headers.');
```
czardoz: No throws please :) 

Check out some of the other classes (`Url`) for what to return when an incorrect type is supplied.

lib/collection/header.js
```diff
@@ -190,6 +192,38 @@ _.extend(Header, /** @lends Header */ {
...
+        }
+        return _.reduce(headers, function (accumulator, header) {
```
czardoz: This won't work for a PropertyList

lib/collection/response.js
```diff
@@ -211,6 +211,29 @@ _.extend(Response.prototype, /** @lends Response.prototype */ {
...
+     * Get the response size
+     * @return {Number}
+     * @todo write unit tests
+     */
+    size: function() {
+        var size = 0,
+            contentLength = Header.getHeaderValue(this.headers.members, 'Content-Length');
+
+        // if 'Content-Length' header present
+        if (contentLength) {
```
czardoz: I remember @shamasis saying something about using these, so just tagging him here

lib/util.js
```diff
@@ -4,7 +4,12 @@ var _ = require('lodash').noConflict(),
 
     ASCII_SOURCE = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789',
     ASCII_SOURCE_LENGTH = ASCII_SOURCE.length,
-    EMPTY = '';
+    EMPTY = '',
+    ECMA_BYTE_SIZES = {
```
czardoz: :+1:

lib/collection/header.js
```diff
@@ -190,6 +192,38 @@ _.extend(Header, /** @lends Header */ {
...
+        return header ? header.value : null;
+    },
+
+    /**
+     * Gets size of an Array or {@link PropertyList} of Headers.
+     *
+     * @param {Array|PropertyList<Header>} headers
+     * @returns {Number}
+     */
+    getSize: function(headers) {
```
czardoz: I'd rather like it if this function was called `size` or `sizeOf`, since this is a static method :smile:

lib/util.js
vegetableman: Addressed the issues.  Thank you for reviewing @czardoz.

lib/collection/response.js
```diff
@@ -211,6 +211,29 @@ _.extend(Response.prototype, /** @lends Response.prototype */ {
...
+     * @return {Number}
+     * @todo write unit tests
+     */
+    size: function() {
+        var responseSize = 0,
+            contentLength = Header.headerValue(this.headers, 'Content-Length');
+
+        // if 'Content-Length' header present
+        if (contentLength) {
+            responseSize = _.parseInt(contentLength);
```
shamasis: please do parsesInt with base 10

shamasis: What happens if this parses to NaN?

vegetableman:
Updated the PR. The `size` method now returns `header size` and `body size`.

Few findings on the header size calculation:- 

Firefox `devtools` calculate response header size either through calling `length` on the raw header information also accounting also for the `whitespace` present.
```
HTTP/1.1 200 OK
Server: nginx/1.8.1
Date: Sun, 24 Jul 2016 23:16:35 GMT
Content-Type: application/json; charset=utf-8
Content-Length: 14
Connection: keep-alive
Access-Control-Allow-Origin: 
Access-Control-Allow-Credentials: 
Access-Control-Allow-Methods: 
Access-Control-Allow-Headers: 
Access-Control-Expose-Headers: 
ETag: W/"e-+3Pmmp/QERKiJq3DueGVYg"
Vary: Accept-Encoding
set-cookie: sails.sid=s%3A5KvXh2N_L9GM7L-ZMtv7TCNbiVmNbH2_.aL8xAy5OL9F9vXEhJEIj5%2Br0bdcrpUPGUX4mB7gZ67o; Path=/; HttpOnly

```
Whereas `curl` calculates bytes per line from the above information, yielding almost the same result as `devtools`. Identical size values were observed on Telerik `fiddler`.

Since, the above raw format is not accessible in the current code or through the `request` lib, recreating the above response format through key-value pair of `headers` object would suffice, as could be seen here.
https://github.com/postmanlabs/postman-collection/pull/65/files#diff-530bbdc33e286200044e07549fa0c349R224 

The problem being the following information is not available through the `Header` class to recreate the same (am guessing keys with empty values are excluded?) :-
```
HTTP/1.1 200 OK
Access-Control-Allow-Origin
Access-Control-Allow-Credentials
Access-Control-Allow-Methods
Access-Control-Allow-Headers
Access-Control-Expose-Headers

```

lib/collection/response.js
```diff
@@ -218,22 +218,22 @@ _.extend(Response.prototype, /** @lends Response.prototype */ {
...
-            responseSize = this.stream ? this.stream.byteLength : (this.body && this.body.length || 0);
+            sizeInfo.bodySize = this.stream ? this.stream.byteLength : (this.body && this.body.length || 0);
         }
 
-        // size of header added
-        responseSize += Header.size(this.headers);
-        return responseSize;
+        // size of header is added
+        sizeInfo.headerSize = Header.size(this.headers);
+        return sizeInfo;
```
shamasis: Why not simply return the detected size of header + body? @vegetableman

lib/util.js
```diff
@@ -241,19 +236,12 @@ util = {
...
      */
-    sizeOf: function(data) {
-       if (_.isString(data)) {
-          return data.length * ECMA_BYTE_SIZES.STRING;
-        } else if (_.isBoolean(data)) {
-          return ECMA_BYTE_SIZES.BOOLEAN;
-        } else if (_.isNumber(data)) {
-          return ECMA_BYTE_SIZES.NUMBER;
-        }
+    isNumeric: function (n) {
```
shamasis: Check if there is equivalet lodash function.

Adding new library function must first be checked whether lodash or other existing functions can serve as same.

lib/collection/header.js
```diff
@@ -217,15 +216,15 @@ _.extend(Header, /** @lends Header */ {
...
         if (!_.isArray(headers) && !PropertyList.isPropertyList(headers)) {
-            return headersSize;
+            return 0;
         }
-        headers.each(function(header) {
-            headersSize += (sizeOf(header.key) + sizeOf(header.value))
-        })
-        return headersSize;
+        var raw = ''
+        headers.each(function (header) {
```
shamasis: Also, @czardoz has a PR almost ready to use the header unparse function instead

lib/util.js
vegetableman: There is no equivalent. https://github.com/lodash/lodash/issues/1148#issuecomment-141139153.

lib/collection/response.js
vegetableman: To allow display of the breakdown of different sizes.

vegetableman:
Added tests. Depends on approval of #69.

lib/collection/response.js
```diff
@@ -211,6 +211,30 @@ _.extend(Response.prototype, /** @lends Response.prototype */ {
...
+            sizeInfo.bodySize = _.parseInt(contentLength, 10);
+        }
+        // else size of body is added
+        else {
+            sizeInfo.bodySize = this.stream ? this.stream.byteLength : (this.body && this.body.length || 0);
+        }
+
+        // size of header is added
+        sizeInfo.headerSize = Header.size(this.headers, this.code, this.reason());
+        return sizeInfo;
```
shamasis: Looks good except for the size info property names. 

You are returning a size, so it is redundant to make the internal property suffixed "size".

response.size().body and respose.size().header sounds more natural.

vegetableman: Cool. Updated.

shamasis:
@czardoz merge this when your header PR is merged.

vegetableman:
Mr. @czardoz the time has come to ... merge.

PR #66, opened 2016-07-22
test commit

czardoz:
do not merge


czardoz:
Had created a fork to test out whether travis works well with forks now.

PR #69, opened 2016-07-25
fixed parsing of headers containing colons

czardoz:
Header parsing now happens only at the first occurance of `:`

### DIFF
--- a/lib/collection/header.js
+++ b/lib/collection/header.js
@@ -132,16 +132,21 @@ _.extend(Header, /** @lends Header */ {
 
     /**
      * Parses a single Header.
-     * @param header
+     * @param {String} header
      * @returns {{key: string, value: string}}
      */
     parseSingle: function (header) {
-        var parsed,
+        if (!_.isString(header)) { return { key: '', value: '' }; }
+
+        var index = header.indexOf(':'),
             key,
             value;
-        parsed = header.split(':');
-        key = parsed.shift();
-        value = parsed.join(':');
+
+        (index < 0) && (index = header.length);
+
+        key = header.substr(0, index);
+        value = header.substr(index + 1);
+
         return {
             key: _.trim(key),
             value: _.trim(value)
--- a/lib/collection/header.js
+++ b/lib/collection/header.js
@@ -157,16 +157,18 @@ _.extend(Header, /** @lends Header */ {
      * Stringifies an Array or {@link PropertyList} of Headers into a single string.
      *
      * @param {Array|PropertyList<Header>} headers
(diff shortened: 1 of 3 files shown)


vegetableman:
@czardoz  It fails for the case `Mon, 25 Jul 2016 13:11:41 GMT`.

vegetableman:
Header is passed a single value `Mon, 25 Jul 2016 13:11:41 GMT` instead of a key, value pair on instantiation.

lib/collection/header.js
```diff
@@ -44,12 +44,7 @@ _.inherit((
      */
     Header = function PostmanHeader (options, name) {
         if (_.isString(options)) {
-            options = Header.parseSingle(options);
-        }
-
-        if (options && _.isString(name)) {
-            options.value = options.key;
-            options.key = name;
+            options = _.isString(name) ? { key: name, value: options } : Header.parseSingle(options);
```
shamasis: why this change?

lib/collection/header.js
```diff
@@ -141,11 +136,15 @@ _.extend(Header, /** @lends Header */ {
      * @returns {{key: string, value: string}}
      */
     parseSingle: function (header) {
-        var parsed;
+        var parsed,
+            key,
+            value;
         parsed = header.split(':');
```
shamasis: why not regex split?

shamasis:
Mostly LGTM.

lib/collection/header.js
vegetableman: while running the tests, the parameter `options` is passed as `Mon, 25 Jul 2016 13:11:41 GMT`  and `name` as `Date`. `parseSingle` breaks for date values like `Mon, 25 Jul 2016 13:11:41 GMT` returning the value as instead `Mon, 25 Jul 2016 13`

lib/collection/header.js
czardoz: Mostly because I find this simpler :-)

shamasis: Then use indexOf and substr if string. They're low level and faster.

test/functional/header.test.js
```diff
@@ -50,6 +50,34 @@ describe('Header', function () {
...
+            var raw = 'name1: value1\r\nname2: value2',
+                list = new PropertyList(Header, {}, raw);
+            expect(list.all()).to.eql([{
+                key: 'name1',
+                value: 'value1'
+            }, {
+                key: 'name2',
+                value: 'value2'
+            }]);
+            expect(Header.unparse(list)).to.eql('name1: value1\nname2: value2')
```
elssar: `to.be` instead of `to.eql`

PR #70, opened 2016-07-27
Added poly chain variable resolution capability, associated tests

kunagpal:
@czardoz


czardoz:
I'm :+1: on this change

shamasis:
The change is non destructive and is good to go. Whatever did not work earlier will not work in any case. In my opinion it does not add new issues. 

Merging.

PR #71, opened 2016-07-28
ensure that we do not throw on url stringification

czardoz:
This bug was causing errors on blank URLs in the App.


shamasis:
LGTM

PR #72, opened 2016-07-28
Return total on response size

vegetableman:
- Handles size for Informational status codes as well.

### DIFF
--- a/lib/collection/header.js
+++ b/lib/collection/header.js
@@ -229,7 +229,7 @@ _.extend(Header, /** @lends Header */ {
         // status-line = HTTP-version SP status-code SP reason-phrase CRLF
         var raw = 'HTTP/X.X' + ' ' + statusCode + ' ' + reason + '\r\n';
         raw += this.unparse(headers, '\r\n');
-        raw += headers.count() ? '\r\n\r\n' : '\r\n';
+        raw += headers.count && headers.count() || headers.length ? '\r\n\r\n' : '\r\n';
         return raw.length;
     }
 });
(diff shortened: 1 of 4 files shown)


vegetableman:
@shamasis have updated the size function.

lib/collection/response.js
```diff
@@ -229,11 +229,13 @@ _.extend(Response.prototype, /** @lends Response.prototype */ {
         }
         // else size of body is added
         else {
-            sizeInfo.body = this.stream ? this.stream.byteLength : (this.body && this.body.length || 0);
+            sizeInfo.body = this.stream && this.stream.byteLength ? this.stream.byteLength :
```
shamasis: Was expecting the byte length based size computation here.

vegetableman: My bad. Missed these comments. Updated it.

lib/collection/header.js
```diff
@@ -229,7 +229,7 @@ _.extend(Header, /** @lends Header */ {
         // status-line = HTTP-version SP status-code SP reason-phrase CRLF
         var raw = 'HTTP/X.X' + ' ' + statusCode + ' ' + reason + '\r\n';
         raw += this.unparse(headers, '\r\n');
-        raw += '\r\n\r\n';
+        raw += headers.count() ? '\r\n\r\n' : '\r\n';
```
shamasis: if headers were ever sent as array, then `.count` would raise errors.

vegetableman: Updated it.

shamasis:
This

PR #73, opened 2016-07-29
do not add existing headers back into the request

czardoz:


shamasis:
Sure? What if it has saved stuff. Should it actually add all heades

shamasis:
AWS auth should set updated headers back overriding.

PR #74, opened 2016-07-31
ensure that the oauth1 signature is encoded if necessary

czardoz:


lib/collection/request-auth/oauth1.js
```diff
@@ -165,6 +165,9 @@ module.exports = {
             accessor.tokenSecret = helperParams.tokenSecret;
         }
         signature = oAuth1.SignatureMethod.sign(message, accessor);
+        if (helperParams.encodeOAuthSign) {
```
shamasis: explanation would help... either here, better if it was in code.

PR #75, opened 2016-07-31
ensure that we don't sign unneeded headers

czardoz:


shamasis:
Does the exclusion of these headers affect AWS auth? Why cache control and not other headers like connection and upgrade?

czardoz:
These are the only two headers added by the app based on user configuration.  In order for AWS Auth to succeed, they should not be forwarded to the aws auth library.

The app currently works around this by adding these headers after processing the Auth

shamasis:
That's okay. What I am wondering is does AWS auth protected servers get affected because these headers are excluded from computation?

For example, to any AWS server, the postman-token is like any other headers.

shamasis:
Next question is that then what are the other headers that AWS auth does NOT use during signature calculation? They all should be removed then.

czardoz:
Yes, there's no such list unfortunately. 
For most services, the only headers that need to be signed are `x-amz-date` and `host`.

PR #78, opened 2016-08-03
added ability to store client side ssl certs in the request object

czardoz:


lib/collection/request.js
```diff
@@ -56,7 +56,15 @@ _.inherit((
...
-            auth: _.createDefined(options, 'auth', RequestAuth)
+            auth: _.createDefined(options, 'auth', RequestAuth),
+
+            /**
+             * @type {Object}
+             * @property {String} certificate
+             * @property {String} key
+             * @todo should this be a separate Property like request body?
+             */
+            ssl: _.isObject(options.ssl) ? options.ssl : {}
```
shamasis: _postman_ssl

shamasis:
Not following this route as per discussion with @czardoz

PR #79, opened 2016-08-03
Updated to use mime-format for guessing mime type

shamasis:
- response.mime() now returns content type detected base after a lot of sanitisation

> note that the structure of mime()-> return object has changed. It now returns 

mime() now returns
```
source: string // 'header', 'content', 'default' or 'forced'
type: string, // sanitised mime type base from mime-fromats module
format: string, // secondary format helper from mime-format module
name: DEFAULT_RESPONSE_FILENAME, // @todo - get from disposition
ext: string // file extension from sanitised content type using mimeType.extension module
filename: string // name + ext,

// also storing some meta info for possible debugging
_originalContentType: string, // the user provided mime type
_sanitisedContentType: string, // sanitised mime type as done by mime-formats module
_accuratelyDetected: boolean // this being true implies worse case (raw render)
detected: {}|null // same as root object, but based on what is detected from content
```

@numaanashraf

### DIFF
--- a/lib/collection/response.js
+++ b/lib/collection/response.js
@@ -153,11 +153,15 @@ _.extend(Response.prototype, /** @lends Response.prototype */ {
      * @note example object returned
      * {
      *   source: string // 'header', 'content', 'default' or 'forced'
-     *   type: string // the content type of response
-     *   ext: string // extension of file that stores this type of data
-     *   name: string // file name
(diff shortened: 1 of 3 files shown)


shamasis:
@czardoz - @numaanashraf gave LGTM. I have also added tests. Can you release and push a patch so that runtime can use it and @numaanashraf can start testing it as soon as he can?

PR #80, opened 2016-08-04
ensure that isEmpty checks in the file content as well

czardoz:
Allows uploading of binary data correctly.

### DIFF
--- a/test/functional/request-body.test.js
+++ b/test/functional/request-body.test.js
@@ -88,6 +88,11 @@ describe('RequestBody', function () {
             expect(body.isEmpty()).to.be(false);
         });
 
+        it('should return false if mode is file and file src is available', function () {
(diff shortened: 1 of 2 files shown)


elssar:
Needs a test case for when both `file.ref` and `file.content` are empty

test/functional/request-body.test.js
```diff
@@ -83,6 +83,27 @@ describe('RequestBody', function () {
             expect(body.isEmpty()).to.be(false);
         });
 
+        it('should return false if mode is file and file src is available', function () {
+            var body = new RequestBody({ mode: 'file', file: { src: '/somewhere/file.txt' } });
+            expect(body.isEmpty()).to.be(false);
+        });
+
+        it('should return false if mode is file and file src is available', function () {
```
elssar: return true if ...

elssar:
LGTM

PR #87, opened 2016-08-18
meta

czardoz:


godfrzero:
Looks good to me.

lib/collection/property-base.js
```diff
@@ -91,6 +91,15 @@ _.extend(PropertyBase.prototype, /** @lends PropertyBase.prototype */ {
...
         }, {});
+    },
+
+    /**
+     * Returns the meta keys associated with the property
+     *
+     * @returns {*}
+     */
+    meta: function () {
+        return arguments.length ? _.pick(this._, Array.prototype.slice.apply(arguments)) : _.cloneDeep(this._);
```
shamasis: It the _.pick will not clone deep. Also I think lodash handles arguments variable directly.

PR #93, opened 2016-08-23
Relaxed response.json() parsing

shamasis:
This PR adds the following

- exposes `util` for testing
- adds `parseJSON` util function that removes BOM and strips comments
- has `strict` flag to turn these off
- adds tests for BOM

TBD: 

1. stripping comment is a good feature?
2. trim the text or strip only BOM?

### DIFF
--- a/test/functional/response.test.js
+++ b/test/functional/response.test.js
@@ -51,6 +51,28 @@ describe('Response', function () {
             });
         });
 
+        it('should strip BOM from response while parsing JSON', function () {
+            expect((new Response({
+                body: String.fromCharCode(0xFEFF) + '{ \"hello\": \"world\" }'
+            })).json()).to.eql({
+                hello: 'world'
+            });
+
+            expect((new Response({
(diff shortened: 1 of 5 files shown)


test/functional/response.test.js
```diff
@@ -51,6 +51,28 @@ describe('Response', function () {
...
+            });
+
+            expect((new Response({
+                body: String.fromCharCode(0xEFBBBF) + '{ \"hello\": \"world\" }'
+            })).json()).to.eql({
+                hello: 'world'
+            });
+        });
+
+        it.only('should strip comments from response while parsing JSON', function () {
```
czardoz: Oops, you did a me!

czardoz:
Should we strip out JSON comments? I'd rather not. It's technically not valid JSON, and I think we should much rather fail on it.

shamasis:
Cool. Changing to the following:

1. Set strict to true by default
2. Checking BOM and showing more detailed error if strict is true
3. removing comment stripping

PR #97, opened 2016-08-24
Strip BOM from JSON parsing of response.json

shamasis:
- get rid of parse-json
- use liquid-json instead
- updated tests


czardoz:
`liquid-json` has no tests. As long are you're sure it works perfectly, I'm :+1: on this.

czardoz:
LGTM

shamasis:
Most tests added by @kunagpal on liquid-json has passed. Merging this.

PR #103, opened 2016-09-08
Added dependency status badges

kunagpal:


czardoz:
I don't want to have the dependency badges, they don't tell people anything significant. Thoughts, @shamasis

shamasis:
we do not want to add such external badges

PR #104, opened 2016-09-08
AWSv4 session token support

kesor:
The AWS library used supports temporary IAM credentials that require the additional parameter of `sessionToken` (https://github.com/mhart/aws4).

This PR solves https://github.com/postmanlabs/newman/issues/353 and https://github.com/postmanlabs/postman-app-support/issues/1663 where adding the session token manually in a header is being handled incorrectly by newman/postman/aws4 library.

Now instead of doing something like this (that doesnt work):
```
"request": {
         "auth": {
                 "type": "awsv4",
                 "awsv4": {
                         "accessKey": "{{accessKey}}",
                         "secretKey": "{{secretKey}}",
                         "region": "{{awsRegion}}",
                         "service": "execute-api",
                         "saveToRequest": true
                 }
         },
         "header": [
                 { "key": "Content-Type", "value": "application/json" },
                 { "key": "X-Amz-Security-Token", "value": "{{sessionToken}}"
         ],
```

you can do something like this, that does work:

```
"request": {
         "auth": {
                 "type": "awsv4",
                 "awsv4": {
                         "accessKey": "{{accessKey}}",
                         "secretKey": "{{secretKey}}",
                         "sessionToken": "{{sessionToken}}",
                         "region": "{{awsRegion}}",
                         "service": "execute-api",
                         "saveToRequest": false
                 }
         },
         "header": [
                 { "key": "Content-Type", "value": "application/json" }
         ],
```

### DIFF
--- a/lib/collection/request-auth/awsv4.js
+++ b/lib/collection/request-auth/awsv4.js
@@ -14,6 +14,7 @@ module.exports = {
         _.extend(this, {
             accessKey: options.accessKey,
             secretKey: options.secretKey,
+            sessionToken: options.sessionToken,
             region: options.region,
             service: options.service
         });
@@ -50,7 +51,8 @@ module.exports = {
         signedData = self.sign({
             credentials: {
                 accessKeyId: params.accessKey,
-                secretAccessKey: params.secretKey
+                secretAccessKey: params.secretKey,
+                sessionToken: params.sessionToken
             },
             host: request.url.getHost(),
             path: request.url.getPathWithQuery(),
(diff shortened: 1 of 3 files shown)


czardoz:
Thanks a lot, @kesor !

shamasis:
Quick question - we did not have this and some services were working okay. Now with this, will that be affected? Point is - if the security token is not provided in options, are we supposed to not send it at all?

kesor:
As far as I tested, should you not provide the extra parameter the AWS library handles that as a nil and works okay. If you add the X-Amz-Security-Token to headers as before AND add the sessionToken, I expect things not to work just like they didn't before.

kesor:
https://github.com/mhart/aws4#aws4signrequestoptions-credentials states the sessionToken parameter is optional, and any falsey value would work according to the code at https://github.com/mhart/aws4/blob/master/aws4.js#L95

czardoz:
Next step is to expose this in app

kesor:
@czardoz Sure. LMK

PR #109, opened 2016-09-12
VariableScope object (and other performance improvements)

shamasis:
This PR does a lot of fine tuning to the SDK and adds a top-level VariableScope class to handle environments and globals.

The generic type of changes:
- performance improvements by using constants
- changes where undefined objects where being saved even when definition did not have them
- made the list indexing functions not hard-code the indexing property in list's functions

This PR is best reviewed at a commit level


lib/collection/variable-scope.js
```diff
@@ -0,0 +1,130 @@
...
+        var values;
+
+        // in case of empty definition simply create values and exit
+        if (!(definition && definition.values)) {
+            values = new VariableList(this);
+        }
+        // set the parent of the externally passed list to match the current
+        else if (VariableList.isVariableList(definition.values)) {
+            if (definition.values.__parent) { // cannot attach a list already part of another collection
+                throw new Error('collection: unable to set a variable-list that already belongs to another scope.');
```
czardoz: Can we clone and use it instead of throwing here?

lib/collection/variable-list.js
```diff
@@ -161,14 +219,16 @@ _.extend(VariableList, /** @lends VariableList */ {
...
+     * case senitivity
      */
     listify: function (items, parent) {
         // if item is plain object transform to array of objects
         _.isPlainObject(items) && (items = _.values(_.mapValues(items, function (value, key) {
-            return {
-                id: key,
-                value: value
-            };
+            var obj = { values: value };
```
czardoz: Earlier, we used to return an object something like this:
```
{ id: 'abc', value: 'def' }
```

With this change, we're now returning 
```
{ id: 'abc', values: 'def' }
```
I don't think this change is correct.

lib/collection/variable-list.js
```diff
@@ -180,19 +240,32 @@ _.extend(VariableList, /** @lends VariableList */ {
...
+                return item;
+            }
+
+            // if the item is a valid object, we use the entire object as a variable definition
+            // also works to handle array type definition along with object type definition
+            var obj = item;
+
+            // in case item is not a valid object, we create our own object + it handles object-type definition
+            if (!(_.isPlainObject(obj) && _.has(obj, 'value'))) {
+                obj = { value: item };
```
czardoz: This will result in an object like
```
{
  id: 'abc'
  value: {
    id: 'abc',
    value: 'def'
  }
}
```
Which does not seem correct...

lib/collection/variable-scope.js
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Convert object to JSON using the base converter
+     * @private
+     * @param {Object} obj
+     * @returns {Object}
+     *
+     * @todo there are more than one instance of this function, maybe move to util or static to Property?
+     */
+    toJSON = function (obj) {
+        return PropertyBase.prototype.toJSON.call(obj);
```
czardoz: Why do we need to explicitly call `PropertyBase.prototype.toJSON`?

lib/collection/variable-list.js
shamasis: This is a bug. Apologies. addressing it

lib/collection/variable-list.js
shamasis: There is a negation in the if block. If the said input is not a plain object or does not have value property, what you mentioned is right.

shamasis: I am thinking of removing the has `value` check.

lib/collection/variable-scope.js
shamasis: I thought of it, but that would result in unexpected behaviour. Imagine you add a variable list by reference and unexpectedly, it has been cloned. This is a usage level edge-case but could have impact. I am cloning it for now and adding an `@note` in docs. okay?

lib/collection/variable-list.js
shamasis: Great great find @czardoz - one reason why I am so confident after you do PR. 😄

Issue #89, opened 2016-08-18
Upgrade to lodash 4.x


shamasis:
Will also cause #88 to go through.
Migration is needed as 4.x is not fully compatible with 3.x
https://github.com/lodash/lodash/wiki/Migrating

PR #110, opened 2016-09-12
Bumped Lodash to v4

kunagpal:
Takes care of #89 

* Replaced JSHint and JSCS with ESLint and it's plugins
* Replaced deprecated method calls from Lodash v3 by their v4 equivalents
* Updated test config to work with ESLint
* Minor code style changes across the project


lib/collection/cookie.js
```diff
@@ -267,11 +270,11 @@ _.extend(Cookie, /** @lends Cookie */ {
         key = split[0].trim();
         value = (split[1]) ? split[1].trim() : true;
 
-        if (_.isString(value) && '"' === value[0]) {
+        if (_.isString(value) && value[0] === QUOT) {
```
czardoz: Can we call it "QUOTE"?

lib/collection/cookie.js
```diff
@@ -267,11 +270,11 @@ _.extend(Cookie, /** @lends Cookie */ {
...
             value = value.slice(1, -1);
         }
 
-        return { key: key, value: value };
+        return {key: key, value: value};
```
czardoz: @shamasis, is this the new ESLint configuration? I rather like the spaces there.

lib/collection/property-list.js
```diff
@@ -155,34 +160,35 @@ _.extend(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
-        var match; // to be used if predicate is an ID
+        var match, // to be used if predicate is an ID
+            list = this;
 
-        !context && (context = this);
+        !context && (context = list);
         // if predicate is id, then create a function to remove that from array
         if (_.isString(predicate)) {
-            (match = this.one(predicate)) && (predicate = function (item) {
+            (match = list.one(predicate)) && (predicate = function (item) {
```
czardoz: Like we discussed the other day, this pattern can cause potential memory leaks. While it's not better or worse than a `.bind()`, there is no reason to change this here.

lib/collection/property-list.js
```diff
@@ -155,34 +160,35 @@ _.extend(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
-                    delete this.reference[index];
+                if ((index = item[list._postman_propertyIndexKey]) && (index = String(index))) {
+                    list._postman_propertyIndexCaseInsensitive && (index = index.toLowerCase());
+                    delete list.reference[index];
                 }
                 delete item[__PARENT]; // unlink from its parent
                 return true;
             }
-        }, this);
+        });
```
czardoz: Use a `.bind()` at this place. That is what the old code used to do, and there's no reason to change it now.

lib/collection/property-list.js
```diff
@@ -229,25 +235,29 @@ _.extend(PropertyList.prototype, /** @lends PropertyList.prototype */ {
      * Iterate on each item of this list
      */
     each: function (iterator, context) {
-        _.each(this.members, iterator, context || this.__parent);
+        _.forEach(this.members, iterator.bind(context || this));
```
czardoz: You're not binding the correct thing here. Look at the diff again.

lib/collection/property-list.js
```diff
@@ -229,25 +235,29 @@ _.extend(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
     },
 
     /**
-     * @param {Function} rule
+     * @param {Function} iterator
      */
-    filter: function (rule, context) {
-        return _.filter(this.members, rule, context);
+    filter: function (iterator, context) {
+        return _.filter(this.members, iterator.bind(context || this));
```
czardoz: Do not introduce changes, the old code does not do a `||`.

lib/collection/property-list.js
```diff
@@ -229,25 +235,29 @@ _.extend(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
      * Find an item within the item group
      *
-     * @param {Function} rule
+     * @param {Function} iterator
      * @param {Object} [context]
      * @returns {Item|ItemGroup}
      */
-    find: function (rule, context) {
-        return _.find(this.members, rule, context);
+    find: function (iterator, context) {
```
czardoz: Do not introduce changes, the old code does not do a `||` for context binding, and neither does it check whether the argument is a function.

lib/collection/property-list.js
```diff
@@ -257,8 +267,7 @@ _.extend(PropertyList.prototype, /** @lends PropertyList.prototype */ {
      * @param context Optional context, defaults to the PropertyList itself.
      */
     map: function (iterator, context) {
-        !context && (context = this);
-        return _.map(this.members, iterator, context);
+        return _.map(this.members, iterator.bind(context || this));
```
czardoz: No `||`.

lib/collection/property.js
```diff
@@ -52,8 +53,9 @@ _.inherit((
              * @note The property can also be present in the `postman_id` meta in case it is not specified in the
              * object. An auto-generated property is used wherever one is not specified
              */
-            id: src.id || (this._ ? this._['postman_id'] : undefined),
-            /*jshint +W069 */
+            id: src.id || (this._ ? this._postman_id : undefined),
```
czardoz: Why is this changed? /cc @shamasis

lib/collection/property.js
```diff
@@ -172,6 +175,7 @@ _.extend(Property, /** @lends Property */ {
      * @returns {Object}
      */
     replaceSubstitutionsIn: function (obj, variables, mutate) {
+        var self = this;
```
czardoz: Use a `.bind()`, since that's the old behavior.

lib/collection/property.js
```diff
@@ -180,9 +184,17 @@ _.extend(Property, /** @lends Property */ {
         // convert the variables to a substitutor object (will not reconvert if already substitutor)
         variables = Substitutor.box(variables, Substitutor.DEFAULT_VARS);
 
-        return _.merge(mutate ? obj : {}, obj, function (objectValue, sourceValue) {
-            return _.isString(sourceValue) ?  this.replaceSubstitutions(sourceValue, variables) : undefined;
-        }, this);
+        return _.mergeWith(mutate ? obj : {}, obj, function customizer(objectValue, sourceValue) {
+            if (!_.isString(sourceValue)) {
```
czardoz: Can you add a test case for this with a very deeply nested object containing a mixture of Arrays and other objects?

lib/collection/query-param.js
```diff
@@ -136,7 +138,7 @@ _.extend(QueryParam, /** @lends QueryParam */ {
         if (!obj) { return ''; }
         var unparsed = obj.key || '';
         // if the value is not null, then we add the value
-        (obj.value != null) && (unparsed += ('=' + (obj.value || '')));
+        (obj.value !== null) && (obj.value !== undefined) && (unparsed += ('=' + (obj.value || '')));
```
czardoz: No changes in behavior please. Add a `todo` if you think this is incorrect, and we'll take it up in a future PR.

lib/collection/request-auth.js
```diff
@@ -207,12 +215,12 @@ _.extend(RequestAuth, /** @lends RequestAuth */ {
             // todo: should we throw an error?
             return request;
         }
-        return handler.authorize.call(handler, request);
+        return handler.authorize(request);
```
czardoz: Why is this changed?

lib/collection/request-auth/oauth1.js
```diff
@@ -103,16 +105,14 @@ module.exports = {
                 system: true
             });
         }
+        else if (_.isString(request.method) && request.method.match(MATCH_PUTPOST) &&
```
czardoz: Not necessary to refactor code here. Revert this, and add a todo if you think this needs a change. We'll take it up in a future PR.

lib/collection/request-auth/oauth1.js
```diff
@@ -103,16 +105,14 @@ module.exports = {
...
+                (request.body && request.body.mode === RequestBody.MODES.urlencoded)) {
+            _.forEach(signatureParams, function (param) {
+                request.body.urlencoded.add(param);
+            });
+        }
         else {
-            if (/PUT|POST/.test(request.method) &&
```
czardoz: Reinstate this code and ensure that the only changes in this file are space removals inside objects.

lib/collection/response.js
```diff
@@ -17,22 +17,27 @@ var util = require('../util'),
...
     DOT = '.',
+
     /**
      * @private
      * @const
      * @type {string}
      */
     DEFAULT_RESPONSE_FILENAME = 'response',
 
+    MATCH_KNOWM_RESPONSE_ENCODING = /^(gzip|deflate)$/,
```
czardoz: Typo in the variable name.

lib/collection/response.js
```diff
@@ -170,14 +177,15 @@ _.extend(Response.prototype, /** @lends Response.prototype */ {
      * @todo write unit tests
      */
     mime: function (contentType, contentDisposition) {
-        var detected = fileType(this.stream || this.body), // detect the mime from response body
+        var self = this,
```
czardoz: Unnecessary

lib/collection/response.js
```diff
@@ -170,14 +177,15 @@ _.extend(Response.prototype, /** @lends Response.prototype */ {
...
+            detected = fileType(self.stream || self.body), // detect the mime from response body
             source = 'forced',
             mime;
 
         // if no overrides provided, we take the values from headers
-        !contentDisposition && (contentDisposition = this.headers.one('content-disposition'));
+        !contentDisposition && (contentDisposition = self.headers.one('content-disposition'));
```
czardoz: No need for `self` here

lib/collection/response.js
```diff
@@ -215,8 +223,8 @@ _.extend(Response.prototype, /** @lends Response.prototype */ {
         }
 
         // we create the body string first from stream and then fallback to body
-        return 'data:' + mime._sanitisedContentType + ';base64, ' + (((this.stream != null) &&
-            util.bufferOrArrayBufferToBase64(this.stream)) || ((this.body != null) && util.btoa(this.body)) || E);
```
czardoz: Is this necessary?

lib/collection/response.js
```diff
@@ -230,7 +238,8 @@ _.extend(Response.prototype, /** @lends Response.prototype */ {
             contentLength = Header.headerValue(this.headers, 'Content-Length');
 
         // if 'Content-Length' header is present and encoding is of type gzip/deflate
-        if (contentLength && /^(gzip|deflate)$/.test(contentEncoding) && util.isNumeric(contentLength)) {
+        if (contentLength && _.isString(contentEncoding) && contentEncoding.match(MATCH_KNOWM_RESPONSE_ENCODING) &&
```
czardoz: Typo in the variable name

lib/collection/url.js
```diff
@@ -144,36 +152,36 @@ _.extend(Url.prototype, /** @lends Url.prototype */ {
...
         }
 
         if (this.host) {
-            rawUrl = rawUrl + this.getHost();
+            rawUrl += this.getHost();
         }
 
         if (this.port) {
-            rawUrl = rawUrl + ':' + this.port.toString();
+            rawUrl += (':' + this.port.toString(options));
```
czardoz: Where did the options come from?

lib/collection/url.js
```diff
@@ -208,8 +216,10 @@ _.extend(Url.prototype, /** @lends Url.prototype */ {
      * @returns {String}
      */
     getQueryString: function (options) {
-        if (this.query.count()) {
-            return QueryParam.unparse(this.query.all(), options);
+        var self = this;
```
czardoz: Pointless

lib/collection/url.js
```diff
@@ -144,36 +152,36 @@ _.extend(Url.prototype, /** @lends Url.prototype */ {
      *
      * @returns {string}
      */
-    toString: function () {
+    toString: function (options) {
```
czardoz: Revert all changes in this function, they're not needed.

lib/collection/request-auth.js
shamasis: This was a redundant .call. Eslint recommended this change afaik.

lib/collection/query-param.js
shamasis: This is same behaviour na? Except that eslint did not suggest to use !=

lib/collection/cookie.js
shamasis: Yes. I felt icky too and then was wondering that this was present in other places. I agree that we should go to one with spaces.

lib/collection/property-list.js
shamasis: This is okay. I am not sure the old code was correct. We should keep the old code and add add a todo saying to check whether it's a bug in old code.

lib/collection/property-list.js
shamasis: This change should ideally not cause any issue. What I would rather love is that code is modified to not do binding at all in case context is falsey. That should make things performant.

lib/collection/property.js
shamasis: Good find. Seems that this is a bug introduced. It appears that the reference branch commit has been directly merged instead of being used as a reference.

lib/collection/request-auth/oauth1.js
shamasis: This was marked as dead block by eslint. And it seems rightly so.

lib/collection/response.js
shamasis: 👍🏼

lib/collection/response.js
shamasis: Eslint recommendation to not use !=. I always felt that != null should be allowed. 😕

PR #112, opened 2016-09-13
Added `hawk` to GreenKeeper ignore

kunagpal:


shamasis:
@czardoz - why are we ignoring hawk updates?

shamasis:
Exactly the reason why we need PRs to have extended description

kunagpal:
@shamasis `hawk` versions beyond the current one `v3.1.3` introduce a breaking change, to which @czardoz can add more context. We end up rejecting Greenkeeper pull requests related to `hawk` anyway, as builds keep breaking.

PR #113, opened 2016-09-15
Feature/oauth1 empty parameters

czardoz:
The parameter that controls the behavior of empty parameters in OAuth 1.0 signing was not being respected previously. Now it is.


czardoz:
Companion changes in `node-oauth1`: https://github.com/postmanlabs/node-oauth1/commit/6bc1f3b544d74dbcbfe284a4ef7982da4171709f

@shamasis

SamvelRaja:
@czardoz Can we have tests for this?

czardoz:
@SamvelRaja There's no easy way to verify generated OAuth signatures, so we rely on integration tests in the app and Newman.

This change works with our OAuth1 endpoint (on echo) and twitter OAuth.

PR #114, opened 2016-09-15
Added leviathan JSON test

kunagpal:


test/functional/complex-property.test.js
```diff
@@ -0,0 +1,72 @@
+var expect = require('expect.js'),
```
czardoz: Please move this inside `property.test.js`

kunagpal: Okay

PR #116, opened 2016-09-16
ensure that security token is removed from headers

czardoz:


shamasis:
addresses: https://github.com/postmanlabs/postman-app-support/issues/1663

PR #147, opened 2016-10-04
Feature/cookie

czardoz:
Corrected the Cookie property, so that it is compliant with https://github.com/salesforce/tough-cookie#properties


czardoz:
@shamasis, will open a separate PR for CookieJars.

czardoz:
Indexing property should be `key`.

PR #148, opened 2016-10-04
Turned off the unit test cobertura reporter on CI.

shamasis:
This would now all now to see full build report on Travis


kunagpal:
@shamasis Shouldn't the Cobertura reporter block have been removed in it's entirety, instead of being commented out?

czardoz:
@kunagpal In case someone uses this repo as a template for other libraries, that commented block will help.

kunagpal:
@czardoz Fair enough. LGTM.

PR #151, opened 2016-10-05
handle empty session token in aws auth

czardoz:


lib/collection/request-auth/awsv4.js
```diff
@@ -53,7 +53,7 @@ module.exports = {
             credentials: {
                 accessKeyId: params.accessKey,
                 secretAccessKey: params.secretKey,
-                sessionToken: params.sessionToken
```
shamasis: Haha! This is such an I told you so moment for me! https://github.com/postmanlabs/postman-collection/pull/104#issuecomment-245868665

PR #152, opened 2016-10-06
implemented a cookie jar

czardoz:


.eslintrc
```diff
@@ -266,7 +266,7 @@
     "lodash/no-double-unwrap": 2,
     "lodash/no-extra-args": 2,
     "lodash/no-single-chain": 2,
-    "lodash/path-style": [2, "string"],
+    // "lodash/path-style": [2, "string"],
```
czardoz: That ensures that I can do:
```
d = _.get(something, [a, b, c]);
```
Which is cleaner than
```
d = something[a] && something[a][b] && something[a][b][c];
```

PR #176, opened 2016-10-18
Proxy support added in the sdk level

SamvelRaja:
With this PR we are now supporting the proxy in the sdk,

The proxy would be a simple object similar to this.
```
{
            url: 'http://example.com',
            protocol: 'http',
            host: 'proxy.com',
            port: '8080',
            tunnel: true
}
```

And this can be achieved in two ways

By providing the object directly into the constructor
```
var Proxy =  require('postman-collection').Proxy,
    p = new Proxy({
      url: 'http://getpostman.com',
      host: 'proxy.com'
    });
```
**PostmanProxy**
```
 {
  url: 'http://getpostman.com',
  protocol: 'http',
  host: 'proxy.com',
  port: '80',
  tunnel: true
}
```

By providing the proxy string to the constructor
```
var Proxy =  require('postman-collection').Proxy,
    p = new Proxy('PROXY [https://proxy.com]:8443', 'gmail.com');
```
**PostmanProxy**
```
{
  url: 'gmail.com',
  protocol: 'https',
  host: 'proxy.com',
  port: '8443',
  tunnel: true 
}

```

### DIFF
--- a/lib/collection/proxy.js
+++ b/lib/collection/proxy.js
@@ -1,0 +1,206 @@
+var _ = require('../util').lodash,
+    Property = require('./property').Property,
+    Url = require('./url').Url,
+
+    E = '',
+    SELECTALL = '*',
+    HTTPS = 'https',
+    HTTP = 'http',
+    DEFAULT_PORT = '80',
+    regexes = {
+        proxyStringSanitizer: /(PROXY)|(DIRECT)|\[|\]|\s/g
+    },
+
+    Proxy;
+
+
+/**
+* You can either provide the options object alone to the constructor or the ProxyString and the UrlString
+*
+* The following is the object structure accepted as constructor parameter while calling `new Proxy(...)`. It is
+* also the structure exported when {@link Property#toJSON} or {@link Property#toObjectResolved} is called on a
+* Proxy instance.
+* @typedef Proxy~definition
+*
+* @property {String=} [url] The url for which the proxy needs to be configured @default would '*' means select all url.
+* @property {String=} [protocol] The proxy server protocol by @default it will be taken as 'http'
+* @property {String=} [host] The proxy server host name @default would be '' [Empty String] represents no proxy
+* @property {String=} [port] The proxy server port @default would be '80'
+* @property {Boolean=} [tunnel] The option to provide whether tunneling needs to be done while proxying the request,
+* @default would be true.
+*
+* @example <caption>JSON definition of an example proxy object</caption>
+* {
+*     url: 'https://example.com',
+*     protocol: 'http',
+*     host: 'proxy.com',
+*     port: 88,
+*     tunnel: true
+* }
+*
+* Incase of ProxyString format
+* @param {String=} [options] supported formats return by the PAC files as provided here
+* (https://web.archive.org/web/20070602031929/http://wp.netscape.com/eng/mozilla/2.0/relnotes/demo/proxy-live.html)
+* @param {String=} [url] The url for which the proxy needs to be configured @default would '*' means select all url.
+
+* @example
+* var proxyObj = new Proxy('PROXY [https://proxy.com]:88', 'http://example.com')
+* {
+*     url: 'http://example.com',
+*     protocol: 'http',
+*     host: 'proxy.com',
+*     port: '88',
+*     tunnel: true
+* }
+*/
+_.inherit((
+    Proxy = function PostmanProxy (options, url) {
+        // this constructor is intended to inherit and as such the super constructor is required to be executed
+        Proxy.super_.call(this, options);
+
+        // Assign defaults before proceeding
+        _.assign(this, /** @lends Variable.prototype */ {
+            /**
+             * The url for which the proxy has been associated with.
+             * @type {String}
+             */
+            url: SELECTALL,
+
+            /**
+             * @type {String}
+             * @default 'http'
+             */
+            protocol: HTTP,
+
+            /**
+             * @type {String}
+             * @default ''
+             */
+            host: E,
+
+            /**
+             * @type {String}
+             * @default '80'
+             */
+            port: DEFAULT_PORT,
+
+            /**
+             * @type {Boolean}
+             * @default true
+             */
+            tunnel: true
+        });
+
+        if (!options) { return; } // in case definition object is missing, there is no point moving forward
+
+        /**
+        * If the options and url is a string then we need to parse the input and update it.
+        */
+        _.isString(options) && (options = Proxy.parse(options, url));
+
+        /**
+        * Make sure the 'url' key property is available in the options provided or generated
+        */
+        this.update(options);
+
+    }), Property);
+
+_.assign(Proxy.prototype, /** @lends Proxy.prototype */ {
+    update: function (options) {
+        if (!_.isObject(options)) {
+            return;
+        }
+        _.assign(this, {
+            /**
+             * The url for which the proxy has been associated with.
+             * @type {String}
+             */
+            url: options.url || SELECTALL,
+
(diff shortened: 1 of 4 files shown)


lib/collection/proxy.js
```diff
@@ -0,0 +1,206 @@
...
+             * @type {String}
+             * @default ''
+             */
+            host: E,
+
+            /**
+             * @type {String}
+             * @default '80'
+             */
+            port: DEFAULT_PORT,
```
czardoz: There is usually no default port for proxies, right? 
If no port is provided, it should be determined from the URL, and set to 443 or 80 depending on whether https is used or not.

Regardless, I think it's best to not assign defaults in the constructor, and assign defaults only in the `.update()` method.

lib/collection/proxy.js
```diff
@@ -0,0 +1,206 @@
...
+    update: function (options) {
+        if (!_.isObject(options)) {
+            return;
+        }
+        _.assign(this, {
+            /**
+             * The url for which the proxy has been associated with.
+             * @type {String}
+             */
+            url: options.url || SELECTALL,
```
czardoz: This should ideally be an instance of `Url`

lib/collection/proxy.js
```diff
@@ -0,0 +1,206 @@
...
+             * @type {String}
+             * @default 80
+             */
+            port: options.port || DEFAULT_PORT,
+
+            /**
+             * @type {Boolean}
+             * @default true
+             */
+            tunnel: options.tunnel || true
```
czardoz: If the URL is `http://` then tunnel should be false, right?

test/integration/proxy.test.js
```diff
@@ -0,0 +1,53 @@
...
+        expect(p.protocol).to.be('http');
+        expect(p.host).to.be('proxy.com');
+        expect(p.port).to.be('8080');
+        expect(p.tunnel).to.be(true);
+    });
+
+    it('Should parse the proxy string provided With DIRECT value where the host must be empty', function () {
+        var p = new Proxy('DIRECT', 'http://example.com');
+
+        expect(p.host).to.be('');
```
czardoz: Is it better to have a separate flag that says direct connections should be used? It seems a little hacky to use use an empty host string to denote direct connections.

lib/collection/proxy.js
```diff
@@ -0,0 +1,206 @@
...
+* {
+*     url: 'http://example.com',
+*     protocol: 'http',
+*     host: 'proxy.com',
+*     port: '88',
+*     tunnel: true
+* }
+*/
+_.inherit((
+    Proxy = function PostmanProxy (options, url) {
```
shamasis: The Proxy constructor documentation is missing.

lib/collection/proxy.js
```diff
@@ -0,0 +1,206 @@
...
+        // this constructor is intended to inherit and as such the super constructor is required to be executed
+        Proxy.super_.call(this, options);
+
+        // Assign defaults before proceeding
+        _.assign(this, /** @lends Variable.prototype */ {
```
shamasis: Do we need such extensive setter in constructor if the update is called anyway?

lib/collection/proxy.js
```diff
@@ -0,0 +1,206 @@
...
+        // If the Url is not available it must be taken as the '*' wild card
+        url = url || SELECTALL;
+
+        /**
+        * proxyString can be DIRECT || PROXY [url]:port
+        * Using the regex /(PROXY)|(DIRECT)|\[|\]|\s/g to replace the unwanted values with ''
+        * For Proxy Load balancing the pac scripts will return the Proxy with two URLs seperated by ';'
+        * We by default will be considering only the first proxy URL
+        */
+        var proxyUrl = proxyString.replace(regexes.proxyStringSanitizer, '').split(';'),
```
shamasis: String constants for E and the ";"

lib/collection/proxy.js
```diff
@@ -0,0 +1,206 @@
...
+            proxyUrlObj = new Url(proxyUrl[0]);
```
shamasis: New line before return. @czardoz why didn't eslint catch this?

lib/collection/proxy.js
```diff
@@ -0,0 +1,206 @@
...
+            /**
+             * @type {Boolean}
+             * @default true
+             */
+            tunnel: true
+        });
+
+        if (!options) { return; } // in case definition object is missing, there is no point moving forward
+
+        /**
```
shamasis: These are inline comments and not jsdoc. Please use normal //

lib/collection/proxy.js
```diff
@@ -0,0 +1,206 @@
...
+        * If the options and url is a string then we need to parse the input and update it.
+        */
+        _.isString(options) && (options = Proxy.parse(options, url));
+
+        /**
```
shamasis: Don't use jsdoc style comment for inline comments.

lib/collection/proxy.js
```diff
@@ -0,0 +1,206 @@
...
+    /**
+     * Specify the key to be used while indexing this object
+     * @private
+     * @readOnly
+     * @type {String}
+     */
+    _postman_propertyIndexKey: 'url',
+
+    /**
+    * Parse the proxy string to postman proxy object
```
shamasis: Jsdoc comment alignment is not correct.

lib/collection/proxy.js
```diff
@@ -0,0 +1,206 @@
+var _ = require('../util').lodash,
+    Property = require('./property').Property,
+    Url = require('./url').Url,
+
+    E = '',
+    SELECTALL = '*',
```
SamvelRaja: 👍

lib/collection/proxy.js
```diff
@@ -0,0 +1,206 @@
...
+    HTTPS = 'https',
+    HTTP = 'http',
+    DEFAULT_PORT = '80',
+    regexes = {
+        proxyStringSanitizer: /(PROXY)|(DIRECT)|\[|\]|\s/g
```
SamvelRaja: 👍

lib/collection/proxy.js
SamvelRaja: yup.. will handle in that way

lib/collection/proxy.js
SamvelRaja: yeah. but as we support wild cards aswell, i don't think so `Url` is the best option for it.

lib/collection/proxy.js
SamvelRaja: Gotcha.

lib/collection/proxy.js
SamvelRaja: 👍

lib/collection/proxy.js
SamvelRaja: 👍

test/integration/proxy.test.js
SamvelRaja: I felt like it is one more props in the proxy, since the DIRECT and host url are related to each other, i find this as a better way.

lib/collection/proxy.js
SamvelRaja: Yeah.. The intention i kept in mind while doing this is, if you want to override all the system proxies, you just need to provide new Proxy(), which will provide you the default values 
```
{
  url: '*',
 protocol: 'http',
 host: '',
 port: '80',
 tunnel: true 
}
```
 which is DIRECT, for all the urls. If want we shall remove this.

test/integration/proxy.test.js
SamvelRaja: Added the direct

PR #180, opened 2016-10-20
Supports ProxyConfig and ProxyConfig List

SamvelRaja:
Proxy Config can be created by 

``` javascript
var ProxyConfig =  require('postman-collection').ProxyConfig,
      p = new ProxyConfig({
                  match: 'http://getpostman.com/*',
                  server: 'htpps://proxy.com',
                  tunnel: true
           });
```

ProxyConfig List is a collection of ProxyConfig

``` javascript
var ProxyConfigList =  require('postman-collection').ProxyConfigList
      list = new ProxyConfigList(parent,
            [
                { match: 'https://example.com*', server: 'https://proxy.com', tunnel: true },
                { match: 'http*://example.com', server: 'https://proxy2.com' },
                { match: 'https://example2*.com', server: 'http://proxyintermediate.com' }
            ]
        );
```

And `list.resolve(url)` will provide you the matched proxy config for the particular url


lib/collection/proxy-config-list.js
```diff
@@ -0,0 +1,80 @@
+var _ = require('../util').lodash,
+    PropertyList = require('./property-list').PropertyList,
+    ProxyConfig = require('./proxy-config').ProxyConfig,
+
+    MATCH_ALL = '*',
+    MATCH_ALL_REGEX = '.*',
```
czardoz: Do you think we should use `.*?` here? `.*` is a greedy operator.
For example, if someone enters a pattern of `http://10.0.*.*/some-path`, the `.*` will match both the asterisks.

If this one, we should probably add a test, because it's very easy to break the behavior with a single character change.

lib/collection/proxy-config.js
```diff
@@ -0,0 +1,114 @@
...
+             * The url mach for which the proxy has been associated with.
+             * @type {String}
+             */
+            match: MATCH_ALL,
+
+            /**
+             * The server url or the proxy url.
+             * @type {Url}
+             */
+            server: new Url(E),
```
czardoz: Just doing a `new Url()` should be okay.

SamvelRaja: Cool

lib/collection/proxy-config-list.js
```diff
@@ -0,0 +1,80 @@
...
+        // Resolving is not needed if the url is not string.
+        if (!_.isString(url)) {
+            return;
+        }
+
+        return this.find(function(proxyConfig) {
+            var matchRegex = new RegExp(PREFIX_DELIMITER +
+                                        proxyConfig.match.replace(MATCH_ALL, MATCH_ALL_REGEX) +
+                                        POSTFIX_DELIMITER);
+            return matchRegex.test(url);
```
shamasis: use `.match` since `.test` is stateful function

lib/collection/proxy-config-list.js
```diff
@@ -0,0 +1,80 @@
...
+        // this constructor is intended to inherit and as such the super constructor is required to be executed
+        ProxyConfigList.super_.call(this, ProxyConfig, parent, populate);
+    }), PropertyList);
+
+_.assign(ProxyConfigList.prototype, /** @lends ProxyConfigList.prototype */ {
+    /**
+     * Matches and gets the proxy config for the particular url
+     * @returns {ProxyConfig~definition=} The matched proxyConfig object
+     * @param {String=} [url] The url for which the proxy config needs to be fetched
+     */
```
shamasis: add a `@todo - use a fixed-length cacheing of regexes in future`

lib/collection/proxy-config.js
```diff
@@ -0,0 +1,114 @@
...
+*
+* @property {String=} [url = *] The match for which the proxy needs to be configured. `*` represents select all.
+* @property {Url=} [server = new Url()] The proxy server url.
+* @property {Boolean=} [tunnel = false] The tunneling option for the proxy request.
+*
+* @example <caption>JSON definition of an example proxy object</caption>
+* {
+*     match: 'https://example.com',
+*     server: 'https://proxy.com',
+*     tunnel: true
```
shamasis: we should add (for sake of example) `disabled: true // to bypass usage of this proxy`

lib/collection/proxy-config.js
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @param {ProxyConfig~definition} options The proxy object structure.
+     */
+    update: function (options) {
+        if (!_.isObject(options)) {
+            return;
+        }
+        _.has(options, 'server') && this.server.update(options.server);
+        _.isString(options.match) && (this.match = options.match);
+        _.has(options, 'server') && (this.tunnel = _.isBoolean(options.tunnel) ? options.tunnel : false);
```
shamasis: 👍

lib/collection/proxy-config.js
```diff
@@ -0,0 +1,230 @@
...
+    },
+
+    /**
+     * Tests the url string with the match provided.
+     * Follows the https://developer.chrome.com/extensions/match_patterns pattern for pattern validation and matching
+     *
+     * @param {String=} [urlStr] The url string for which the proxy match needs to be done.
+     */
+
+    test: function(urlStr) {
```
czardoz: This function needs to be broken down, it's too big right now, and contains a lot of conditionals, which can cause issues with tests.

lib/collection/proxy-config.js
```diff
@@ -0,0 +1,230 @@
...
+        // Check the match pattern of sanity and split it into protocol, host and path
+        match = this.match.match(regexes.patternSplit);
+
+        if (!match) {
+            // This ensures it is a invalid match pattern
+            return false;
+        }
+
+        // If the protocol retuns '*' we are going to assign ['https', 'http'] array to it.
+        matchRegexObject.protocol = match[1] === MATCH_ALL ? ALLOWED_PROTOCOLS : [match[1]];
```
czardoz: you can make a new function that will directly return  a `matchRegexObject`, so that you can test it out separately in the unit tests.

lib/collection/proxy-config.js
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        /*
+        * For Host match, we are considering the port with the host, hence we are using getRemote() instead of getHost()
+        * We need to address three cases for the host urlStr
+        * 1. * It matches all the host + protocol,  hence we are not having any parsing logic for it.
+        * 2. .*foo.bar.com Here the prefix could be anything but it should end with foo.bar.com
+        * 3. foo.bar.com This is the absolute matching needs to done.
+        */
+        if (matchRegexObject.host === MATCH_ALL) {
```
czardoz: This logic should be simplified. There's no need for an empty if block.

lib/collection/proxy-config.js
```diff
@@ -0,0 +1,230 @@
...
+          // Don't check anything in host. simple go ahead for the path matching logic
+        }
+        else if (matchRegexObject.host[0] === MATCH_ALL) {
+            // It must be *.foo. We also need to allow foo by itself.
+            hostSuffix = matchRegexObject.host.substr(2);
+            if (url.getRemote() !== hostSuffix && !url.getRemote().endsWith('.' + hostSuffix)) {
```
czardoz: Since you are returning in this block, you should club it together into a single `if`

test/integration/proxy-config.test.js
```diff
@@ -0,0 +1,36 @@
+var expect = require('expect.js'),
+    ProxyConfig = require('../../lib/index.js').ProxyConfig;
+
+/* global describe, it */
+describe('Proxy Config', function () {
```
czardoz: Should add a lot of tests for the `ProxyConfig.test` method specifically. In fact, we should have 100% branch coverage for that method.

SamvelRaja: Added the unit tests and it is 100% coverage now 😄

lib/collection/proxy-config.js
SamvelRaja: This function has been broken down into util functions.

PR #185, opened 2016-10-24
The proxy list matching logic has been revisited and standardised

SamvelRaja:
The proxy config resolver has been changes by adding the match pattern standardized according to the chrome [match pattern](https://developer.chrome.com/extensions/match_patterns) spec.

### DIFF
--- a/lib/collection/proxy-config.js
+++ b/lib/collection/proxy-config.js
@@ -2,8 +2,22 @@ var _ = require('../util').lodash,
     Property = require('./property').Property,
     Url = require('./url').Url,
 
-    E = '',
     MATCH_ALL = '*',
+    PREFIX_DELIMITER = '^',
+    POSTFIX_DELIMITER = '$',
+    MATCH_ALL_URLS = '<all_urls>',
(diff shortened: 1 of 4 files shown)


lib/collection/proxy-config.js
```diff
@@ -90,6 +104,105 @@ _.assign(ProxyConfig.prototype, /** @lends ProxyConfig.prototype */ {
...
+
+        // If all the conditions are satisfied then it is matched.
+        return true;
+    },
+
+    globPatternToRegexp: function(pattern) {
+        // Escape everything except ? and *.
+        pattern = pattern.replace(regexes.escapeMatcher, regexes.escapeMatchReplacement);
+        pattern = pattern.replace(regexes.questionmarkMatcher, regexes.questionmarkReplacment);
+        pattern = pattern.replace(regexes.starMatcher, regexes.starReplacement);
```
shamasis: We need to also do a sort of escape so that we don't have a security risk from user info.

shamasis:
Also - can you document the function definitions that are not?

PR #192, opened 2016-11-01
Improved response documentation structure and added encoding helper function

shamasis:


shamasis:
@numaanashraf - using this, we can temporarily implement lib decompression in app response if 

```
var enc = response.encoding();

if (enc.format === 'gzip' && enc.source === 'body') {
     response.stream = (new zlib.gunzip()).pipe(response.stream); // pseudo-code
}
```

lib/collection/response.js
```diff
@@ -63,7 +110,7 @@ _.assign(Response.prototype, /** @lends Response.prototype */ {
              * @type {String}
              * @deprecated use .reason()
              */
-            status: options.status ? options.status : httpReasons.lookup(options.code).name,
+            status: _.pluck(options.reason, options.status, httpReasons.lookup(options.code).name),
```
czardoz: `_.pluck` used to be a lodash 3 function which used to do something different. Sure we should use that name?

lib/collection/response.js
```diff
@@ -246,6 +304,38 @@ _.assign(Response.prototype, /** @lends Response.prototype */ {
...
+     * @private
+     * @returns {Object} - {format: string, source: string}
+     */
+    encoding: function () {
+        var contentEncoding = Header.headerValue(this.headers, CONTENT_ENCODING),
+            body = this.stream || this.body,
+            source;
+
+        if (contentEncoding) {
+            source = 'header';
```
czardoz: Probably should be `HEADER`

lib/util.js
```diff
@@ -184,6 +186,14 @@ _.mixin(/** @lends util */ {
             result[i] = ASCII_SOURCE[(Math.random() * ASCII_SOURCE_LENGTH) | 0];
         }
         return result.join(EMPTY);
+    },
+
+    pluck: function () {
```
czardoz: https://github.com/postmanlabs/postman-collection/pull/192/files#r86092342

czardoz:
@kunagpal, this is a simple change with not many comments, can you update the PR?

kunagpal:
@czardoz Alright, I'll get on this.

lib/util.js
kunagpal: How about renaming to pick?

czardoz: `_.choose` sounds closer to the original intended meaning

kunagpal:
@czardoz Are there any other changes needed here?

PR #193, opened 2016-11-01
Remove non-existing directories from `package.json`

amaechler:
Thanks to @jaydeep17 for his comment: [yarnpkg/yarn#739 (comment)](https://github.com/yarnpkg/yarn/issues/739#issuecomment-255598619)

This change allows installing this package (or a parent) with yarn.


czardoz:
Thanks @amaechler! :cake:

PR #216, opened 2016-11-18
Converted `timeStamp` to `timestamp` in oAuth1

SamvelRaja:


czardoz:
@shamasis, released changes in schema & transformer

PR #231, opened 2016-12-06
Clone the request only if the auth is specified in the request

SamvelRaja:


czardoz:
@SamvelRaja I'm closing this as discussed, we can open it again if we run into issues :)

PR #232, opened 2016-12-07
Find the encoding from the header and encode the stream response accordingly

SamvelRaja:
Also closed the issue https://github.com/postmanlabs/postman-app-support/issues/2502


SamvelRaja:
@czardoz Will add tests gimme some time.

lib/util.js
```diff
@@ -222,13 +223,21 @@ util = {
...
-    bufferOrArrayBufferToString: function (buffer) {
+    bufferOrArrayBufferToString: function (buffer, charset) {
         if (!buffer || _.isString(buffer)) {
             return buffer || '';
         }
-
-        var str = buffer.toString('utf8') || '';
-        (str === '[object ArrayBuffer]') && (str = util.arrayBufferToString(buffer));
+        var str = '';
+        try {
```
shamasis: flow incorrect.

this function provides same api for buffer and array buffer. if array buffer is detected first, we operate and bail out,

lib/collection/response.js
```diff
@@ -123,7 +130,7 @@ _.assign(Response.prototype, /** @lends Response.prototype */ {
      * @returns {String|undefined}
      */
     text: function () {
-        return (this.stream ? util.bufferOrArrayBufferToString(this.stream) : this.body);
+        return (this.stream ? util.bufferOrArrayBufferToString(this.stream, this.mime().charset) : this.body);
```
SamvelRaja: @shamasis mime uses **this.stream || this.body**  [here](https://github.com/postmanlabs/postman-collection/pull/232/files#diff-a2d0c3519ebfb4ceb7c93d7911ee0718L178),  so using mime() here is safe

PR #240, opened 2016-12-21
Provided the public api to test the match pattern

SamvelRaja:
@czardoz  provided the way to test the matchpattern without creating the config, which helps in having a button in app to test the match pattern.


lib/collection/proxy-config.js
```diff
@@ -107,6 +107,31 @@ _.assign(ProxyConfig.prototype, /** @lends ProxyConfig.prototype */ {
     },
 
     /**
+     * Used to generate the match regex object from the match string we have.
+     * @returns Match regex object
+     */
+    validateMatchPattern: function(matchPattern) {
```
SamvelRaja: Agreed changed..

lib/collection/proxy-config.js
```diff
@@ -107,6 +107,31 @@ _.assign(ProxyConfig.prototype, /** @lends ProxyConfig.prototype */ {
...
+        // No need to match if it is not a valid string.
+        if (!_.isString(matchPattern) || _.isEmpty(matchPattern)) {
+            return ({});
```
SamvelRaja: @czardoz 
For _non string and empty string matchPattern_, i don't want to go into the **.match** stage which is costlier and anyway we knows that it will fails.
Also, IMO i preferred to have the same return type and can be checked _isEmpty_ for **falseness**.

PR #248, opened 2017-01-13
Ignore dots within variables on host

vegetableman:
This fixes the issue: https://github.com/postmanlabs/postman-app-support/issues/2629


vegetableman:
@shamasis ^

shamasis:
@vegetableman can you add a test for recursive variables? {{an.d-{{ano.ther}}}}

vegetableman:
@czardoz Added more tests.
@shamasis Are we supporting nested variables? don't think a regex could take care of that case. Also, Ideally, the url class shouldn't care about variable resolution. Let me know if there's a better way of doing this fix.

czardoz:
@vegetableman, the URL class is _not_ doing variable resolution, it's doing parsing of URLs containing variables. What @shamasis means is, your regex should ensure that any nested variables are preserved after parsing, meaning it should only consider the outermost `{{`

vegetableman:
@czardoz I meant, is there any way we could do the domain split on the resolved values instead of the raw url itself?

czardoz:
No, we're not trying to resolve anything here, just parsing the domain from the URL. There's no other way to do it :)

vegetableman:
Updated. Couldn't find a place for `Splitter` other than `superstring` (it's not being used for the value though). Let me know if I should move it or refactor it.

vegetableman:
@czardoz @shamasis Simplified the logic to do join after split.

vegetableman:
@czardoz @shamasis Brought it back to it's original glory :/

SamvelRaja:
Fixes the issue.
https://github.com/postmanlabs/postman-app-support/issues/2620

PR #271, opened 2017-02-08
Allow options.stream to be a new Buffer().toJSON()

kamalaknn:


lib/collection/response.js
```diff
@@ -107,6 +107,8 @@ _.inherit((
 
 _.assign(Response.prototype, /** @lends Response.prototype */ {
     update: function (options) {
+        // options.stream accepts new Buffer() as well as new Buffer().toJSON()
+        options.stream && options.stream.type === 'Buffer' && (options.stream = new Buffer(options.stream));
```
czardoz: Newer node versions use `Buffer.from()`, `new Buffer()` is deprecated, so can you handle that case? I think @kunagpal has handled it somewhere already.

Also, is `Buffer` available in browsers? This code is cross platform, so it should work in browsers as well.

SamvelRaja: @czardoz This is the time, shall we stop assigning body to stream and follow, stream to stream and body to body props.

czardoz: @SamvelRaja it's used in Newman & monitoring I think, so let's not do that yet.

SamvelRaja: I didn't see this... if i had, i would pushed in 4.0 itself 😿

shamasis: 1. The change foes inside mergeDefined (L144)
2. In the turnery else block, detect type=Buffer && data=Array then
3. If Buffer present, create Buffer
4. else create ArrayBuffer from data Array (keep performance in mind)

test/unit/response.test.js
```diff
@@ -23,6 +23,16 @@ describe('Response', function () {
...
 
+        it('must retain Buffer nature on stream property after new Response(response.toJSON())', function () {
+            var rawResponse = {
+                    body: 'response body',
+                    stream: new Buffer([114, 101, 115, 112, 111, 110, 115, 101, 32, 98, 111, 100, 121])
+                },
+                response = new Response(rawResponse),
+                jsonified = response.toJSON(),
+                reconstructedResponse = new Response(jsonified);
+            expect(reconstructedResponse.stream.toString()).to.eql(response.stream.toString());
```
shamasis: test on fixture and not on processed response

lib/collection/response.js
```diff
@@ -107,6 +115,8 @@ _.inherit((
 
 _.assign(Response.prototype, /** @lends Response.prototype */ {
     update: function (options) {
+        // options.stream accepts new Buffer() as well as new Buffer().toJSON()
+        options.stream = normaliseStream(options.stream);
```
kamalaknn: @shamasis Doing this here over L144, so that `options.stream` has `byteLength` for `options.stream.byteLength` on L161.

shamasis: No please. You're mutating options object.

lib/collection/response.js
```diff
@@ -11,6 +11,14 @@ var util = require('../util'),
     Cookie = require('./cookie').Cookie,
     PropertyList = require('./property-list').PropertyList,
 
+    normaliseStream = function (stream) {
+        if (stream && stream.type === 'Buffer' && _.isArray(stream.data)) {
+            // @todo Add tests for Browser environments, where ArrayBuffer is returned instead of Buffer
+            return typeof Buffer === 'function' ? new Buffer(stream.data) : new Uint8Array(stream.data).buffer;
```
shamasis: Also, can you double check and confirm Uint8Array().buffer actually works in browsers? Also move the 'function' string to a variable on top.

lib/collection/response.js
```diff
@@ -11,6 +11,14 @@ var util = require('../util'),
     Cookie = require('./cookie').Cookie,
     PropertyList = require('./property-list').PropertyList,
 
+    normaliseStream = function (stream) {
+        if (stream && stream.type === 'Buffer' && _.isArray(stream.data)) {
```
shamasis: Move the string "Buffer" to a constant (var) on top.

kamalaknn:
@shamasis Pushed with the changes. Checked `new UInt8Array().buffer` on browser. Referencing [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/buffer).

test/unit/response.test.js
```diff
@@ -23,6 +24,16 @@ describe('Response', function () {
             expect(jsonified).to.have.property('cookie');
         });
 
+        it('must retain Buffer nature on stream property after new Response(response.toJSON())', function () {
+            var rawResponse = {
+                    body: 'response body',
```
SamvelRaja: remove this body prop.

SamvelRaja: @kamalaknn Also, can you add an explicit test for 
var rawResponse = { 
stream = {
  type: 'Buffer',
  data: [ 114, 101, 115, 112, 111, 110, 115, 101, 32, 98, 111, 100, 121 ]
}
}

I know this covers this usecase, but expected a seperate one as well.

kamalaknn: The reason I did not want to add an explicit test for the above case, is because we are not really interested in supporting a `{ type: 'Buffer', data: [] }`. What is more important, is to ensure `response` and `new Response(response.toJSON())` are the same, i.e. response instance is properly serialised and deserialised back without loosing data. Which is what the added test ensures.

PR #274, opened 2017-02-13
Added .parent helper for items and itemGroups

kunagpal:


kunagpal:
@czardoz Added

test/unit/item-group.test.js
```diff
@@ -14,4 +14,19 @@ describe('ItemGroup', function () {
         });
         expect(groups.length).to.be(4);
     });
+
+    describe('.parentOf checks', function () {
+        it('should work correctly for nested collections', function () {
+            var collection = new Collection(fixtures.nestedCollectionV2);
```
czardoz: Can we not use the fixtures? They're wreaking havoc and we are supposed to get rid of them soon.. every time we add fixtures, someone or the other ends up adding new tests based on them, and then our tests get coupled.

test/unit/item-group.test.js
```diff
@@ -14,4 +14,19 @@ describe('ItemGroup', function () {
...
+
+            collection.forEachItemGroup(function (group) {
+                var parent = group.parentOf();
+
+                while (parent) {
+                    expect(parent).to.have.keys(['description', 'id', 'name', 'items', 'events']);
+                    parent = parent.parentOf();
```
czardoz: Add `parent.id`s to an array, and verify that the order is as expected.

test/unit/item.test.js
```diff
@@ -20,4 +21,21 @@ describe('Item', function () {
...
+
+    describe('.parentOf', function () {
+        it('must return a falsy result for a standalone item', function () {
+            expect(item.parentOf()).to.not.be.ok();
+        });
+
+        it('must correctly return the parent for a provided item', function () {
+            var collection = new Sdk.Collection(fixtures.collectionV2);
+
+            _.forEach(collection.items.members, function (item) {
```
czardoz: Use `collection.forEachItem()`

kunagpal: Done.

test/unit/item-group.test.js
kunagpal: Okay, I'll add a todo here.

test/unit/item.test.js
```diff
@@ -20,4 +20,23 @@ describe('Item', function () {
             expect(jsonified).to.have.property('event');
         });
     });
+
+    describe('.parent', function () {
+        it('must return a falsy result for a standalone item', function () {
+            expect(item.parent()).to.not.be.ok();
+        });
+
+        it('must correctly return the parent for a provided item', function () {
```
czardoz: 4 cases:
```
1. nested item (compare with hardcoded expected value)
2. nested itemgroup (compare with hardcoded expected value)
3. first level item (should return collection as parent)
4. first level itemgroup (should return collection as parent)

kunagpal: Added

PR #283, opened 2017-02-22
PropertyBase#forEachParent

kunagpal:
* Includes options to return the collection instance as well, depending on the value of the `withRoot` option.


lib/collection/property-base.js
```diff
@@ -60,6 +60,28 @@ PropertyBase = function PropertyBase (definition) {
...
+     *
+     * @param {?Object|Boolean} [options={}] - A set of options for the parent chain traversal.
+     * @param {?Boolean} [options.withRoot=false] - Set to true to include the collection object as well.
+     * @param {Function} iterator - The function to call for every parent in the ancestry chain.
+     * @todo Cache the results
+     */
+    forEachParent: function (options, iterator) {
+        _.isFunction(options) && (iterator = options, options = {});
+        if (!_.isFunction(iterator)) { return; }
+        if (!_.isObject(options)) { options = { withRoot: options }; }
```
czardoz: This should not be supported. There should be no calls that look like 
```
item.forEachParent(false, function () {})
```

lib/collection/property-base.js
```diff
@@ -60,6 +60,28 @@ PropertyBase = function PropertyBase (definition) {
...
+
+        var parent = this.parent(),
+            grandparent = parent && parent.parent();
```
czardoz: safety is paramount, we _never_ throw errors in the SDK, so add extra checks to ensure that `.parent` is a function

lib/collection/property-base.js
```diff
@@ -60,6 +60,28 @@ PropertyBase = function PropertyBase (definition) {
...
+
+        while (parent && (grandparent || options.withRoot)) {
+            iterator(parent);
+            (parent = grandparent, grandparent = grandparent && grandparent.parent());
```
czardoz: Unnecessary comma expression. This needs to be separate statements.

test/unit/property-base.test.js
```diff
@@ -1,6 +1,8 @@
 var _ = require('lodash'),
     expect = require('expect.js'),
-    PropertyBase = require('../../lib/index.js').PropertyBase;
+    Sdk = require('../../lib/index.js'),
```
czardoz: We only capitalize class names. Since `Sdk` is not a classname, use lowercase. (Also, if you've done this in the other PR for `.parent`, fix it in this one.

test/unit/property-base.test.js
```diff
@@ -24,11 +26,49 @@ describe('PropertyBase', function () {
...
+
+            item.forEachParent(chain.unshift.bind(chain));
+            expect(_.map(chain, 'name')).to.eql(expectedOrder);
+        });
+
+        it('should return the collection as well when required', function () {
+            var chain = [];
+
+            item.forEachParent({ withRoot: true }, chain.unshift.bind(chain));
+            expect(chain.shift().id).to.be(fixtures.nestedCollectionV2.info._postman_id);
```
czardoz: This kinda thing is why you should use separate `expectedOrder` arrays to test. Needlessly combining the expected order has created this correct, but non-straightforward test.

test/unit/property-base.test.js
```diff
@@ -24,11 +26,33 @@ describe('PropertyBase', function () {
...
+            item.forEachParent(chain.unshift.bind(chain));
+            expect(_.map(chain, 'name')).to.eql(expectedOrder);
+        });
+
+        it('should return the collection as well when required', function () {
+            var chain = [],
+                expectedOrder = ['F2', 'F2.F3'];
+
+            item.forEachParent({ withRoot: true }, chain.unshift.bind(chain));
+            expect(chain.shift().id).to.be(fixtures.nestedCollectionV2.info._postman_id);
```
czardoz: gg

PR #286, opened 2017-02-27
Introduce Certificate and CertificateList

kamalaknn:
- Introduce Certificate and CertificateList
  - Extract url matching logic into UrlMatchPattern and UrlMatchPatternList
  - Remove url matching code from ProxyConfig
  - Restructure tests

### DIFF
--- a/lib/collection/certificate.js
+++ b/lib/collection/certificate.js
@@ -1,0 +1,161 @@
+var _ = require('../util').lodash,
+    Property = require('./property').Property,
+    Url = require('./url').Url,
+    UrlMatchPatternList = require('../url-pattern/url-match-pattern-list').UrlMatchPatternList,
+
+    STRING = 'string',
+    HTTPS = 'https',
+
+    Certificate;
+
+/**
+* The following is the object representation accepted as param for the Certificate constructor.
+* Also the type of the object returned when {@link Property#toJSON} or {@link Property#toObjectResolved} is called on a
+* Certificate instance.
+* @typedef Certificate~definition
+* @property {String} [name] A name for the certificate
+* @property {Array} [matches] A list of match patterns
+* @property {{ src: (String) }} [key] Object with path on the file system for private key file, as src
+* @property {{ src: (String) }} [cert] Object with path on the file system fot certificate file, as src
+* @property {String} [passphrase] The passphrase for the certificate key
+*
+* @example <caption>JSON definition of an example certificate object</caption>
+* {
+*     name: 'My certificate for example.com',
+*     matches: ['https://example.com/*'],
+*     key: { src: '/path/to/key' },
+*     cert: { src: '/User/path/to/certificate' },
+*     passphrase: 'iampassphrase'
+* }
+*/
+_.inherit((
+
+    /**
+     * A Certificate definition that represents the ssl certificate
+     * to be used for an url.
+     * Properties can then use the `.toObjectResolved` function to procure an object representation of the property with
+     * all the variable references replaced by corresponding values.
+     * @constructor
+     * @extends {Property}
+     *
+     * @param {Certificate~definition=} [options] Object with matches, key, cert and passphrase
+     *
+     * @example <caption> Create a new Certificate</caption>
+     *
+     * var Certificate = require('postman-collection').Certificate,
+     *    certificate = new Certificate({
+     *     name: 'Certificate for example.com',
+     *     matches: ['example.com'],
+     *     key: { src: '/User/path/to/certificate/key' },
+     *     cert: { src: '/User/path/to/certificate' },
+     *     passphrase: 'iampassphrase'
+     * });
+     */
+    Certificate = function Certificate (options) {
+        // this constructor is intended to inherit and as such the super constructor is required to be executed
+        Certificate.super_.apply(this, options);
+
+        this.update(options);
+    }), Property);
+
+_.assign(Certificate.prototype, /** @lends Certificate.prototype */ {
+    /**
+     * Ensure all object have id
+     * @private
+     */
+    _postman_requiresId: true,
+
+    /**
+     * Updates the certificate with the given properties
+     * @param {Certificate~definition=} [options] Object with matches, key, cert and passphrase
+     */
+    update: function (options) {
+        // return early if options is empty or invalid
+        if (!_.isObject(options)) {
+            return;
+        }
+
+        _.mergeDefined(this, {
+            /**
+             * Unique identifier
+             * @type {String}
+             */
+            id: options.id,
+
+            /**
+             * Name for user reference
+             * @type {String}
+             */
+            name: options.name,
+
+            /**
+             * List of match pattern
+             * @type {UrlMatchPatternList}
+             */
+            matches: options.matches && new UrlMatchPatternList({}, options.matches),
+
+            /**
+             * Private Key
+             * @type {{ src: (string) }}
+             */
+            key: options.key,
+
+            /**
+             * Certificate
+             * @type {{ src: (string) }}
+             */
+            cert: options.cert,
+
+            /**
+             * passphrase
+             * @type {Object}
+             */
+            passphrase: options.passphrase
+        });
+    },
+
+    /**
+     * Checks if the certificate can be applied to a given url
+     *
+     * @param {String|Url} url The url string for which the certificate is checked for match.
+     */
+    canApplyTo: function(url) {
+        if (_.isEmpty(url)) {
+            return false;
+        }
+
+        // convert url strings to Url
+        (typeof url === STRING) && (url = new Url(url));
+
+        // this ensures we don't proceed any further for any protocol
+        // that is not https
+        if (url.protocol !== HTTPS) {
+            return false;
+        }
+
+        // test the input url against allowed matches
+        return this.matches.test(url);
+    }
+});
+
+_.assign(Certificate, /** @lends Certificate */ {
+    /**
+     * Defines the name of this property for internal use
+     * @private
+     * @readOnly
+     * @type {String}
+     */
+    _postman_propertyName: 'Certificate',
+
+    /**
+     * Specify the key to be used while indexing this object
+     * @private
+     * @readOnly
+     * @type {String}
+     */
+    _postman_propertyIndexKey: 'id'
+});
+
+module.exports = {
+    Certificate: Certificate
+};
--- a/lib/collection/certificate.js
+++ b/lib/collection/certificate.js
@@ -98,13 +98,13 @@ _.assign(Certificate.prototype, /** @lends Certificate.prototype */ {
              * Private Key
              * @type {{ src: (string) }}
              */
-            key: options.key,
+            key: _.isObject(options.key) ? options.key : { src: options.key },
 
             /**
              * Certificate
              * @type {{ src: (string) }}
              */
-            cert: options.cert,
+            cert: _.isObject(options.cert) ? options.cert : { src: options.cert },
 
             /**
              * passphrase
--- a/lib/collection/proxy-config.js
+++ b/lib/collection/proxy-config.js
@@ -1,23 +1,7 @@
 var _ = require('../util').lodash,
     Property = require('./property').Property,
     Url = require('./url').Url,
-
-    MATCH_ALL = '*',
-    PREFIX_DELIMITER = '^',
-    POSTFIX_DELIMITER = '$',
-    MATCH_ALL_URLS = '<all_urls>',
-    ALLOWED_PROTOCOLS = ['http', 'https'],
-    ALLOWED_PROTOCOLS_REGEX = ALLOWED_PROTOCOLS.join('|'),
-
-    regexes = {
-        escapeMatcher: /[.+^${}()|[\]\\]/g,
-        escapeMatchReplacement: '\\$&',
-        questionmarkMatcher: /\?/g,
-        questionmarkReplacment: '.',
-        starMatcher: '*',
-        starReplacement: '.*',
-        patternSplit: '^(' + ALLOWED_PROTOCOLS_REGEX + '|\\*)*://(\\*|\\*\\.[^*/]+|[^*/]+|)(/.*)$'
-    },
+    UrlMatchPattern = require('../url-pattern/url-match-pattern').UrlMatchPattern,
 
     ProxyConfig;
 
@@ -35,7 +19,7 @@ var _ = require('../util').lodash,
 *
 * @example <caption>JSON definition of an example proxy object</caption>
 * {
-*     match: 'https://example.com',
+*     match: 'https://example.com/*',
 *     server: 'https://proxy.com',
 *     tunnel: true,
 *     disabled: false
@@ -73,7 +57,7 @@ _.inherit((
              * The url mach for which the proxy has been associated with.
              * @type {String}
              */
-            match: MATCH_ALL_URLS,
+            match: new UrlMatchPattern(),
 
             /**
              * The server url or the proxy url.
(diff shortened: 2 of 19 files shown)


lib/collection/url-match-pattern-list.js
```diff
@@ -0,0 +1,88 @@
+var _ = require('../util').lodash,
```
czardoz: Ideally, any class that is defined inside `lib/collection` should represent a data structure that exists inside the Postman Collection Format. 

This is mostly a helper class, so it should not be inside `lib/collection`. Can you move this to a new folder, `lib/url-pattern/match-list.js`?

kamalaknn: Agreed 👍

lib/url-pattern/url-match-pattern-list.js
```diff
@@ -0,0 +1,88 @@
...
+
+        matchedSpecificPattern = this.find(function(urlMatchPattern) {
+            var matchRegexObject = urlMatchPattern.createMatchPattern(urlMatchPattern.pattern);
+
+            // Empty matchRegexObject represents the match is INVALID match
+            if (_.isEmpty(matchRegexObject)) {
+                return false;
+            }
+
+            return (urlMatchPattern._testProtocol(matchRegexObject, url.protocol) &&
```
SamvelRaja: don't leak the private functions of urlMatchPattern into urlMatchPatternList,
use 

> urlMatchPattern.test(urlString)

lib/url-pattern/url-match-pattern-list.js
```diff
@@ -0,0 +1,88 @@
...
+         * this method first searches for MATCH_ALL_URLS in all patterns
+         * and then moves on to the slower Regex based searches.
+         */
+        var url,
+            matchAllUrlsPattern,
+            matchedSpecificPattern;
+
+        matchAllUrlsPattern = this.find(function(urlMatchPattern) {
+            return urlMatchPattern.pattern === MATCH_ALL_URLS;
+        });
```
SamvelRaja: Not a big fan for this :(  it anyway will take to an unwanted loop over list, if we have hundreds of matchpattern without _MATCH_ALL_URLS_ IMO

SamvelRaja: when you use **urlMatchPattern.test(urlString)**  that itself will bail out on the first occurrence of **MATCH_ALL_URLS**

lib/collection/proxy-config.js
```diff
@@ -73,7 +59,7 @@ _.inherit((
              * The url mach for which the proxy has been associated with.
              * @type {String}
              */
-            match: MATCH_ALL_URLS,
+            match: new UrlMatchPattern(MATCH_ALL_URLS),
```
SamvelRaja: new UrlMatchPattern() should defaults to MATCH_ALL_URLS instead of getting the value from here.

lib/collection/certificate.js
```diff
@@ -0,0 +1,129 @@
...
+        // this constructor is intended to inherit and as such the super constructor is required to be executed
+        Certificate.super_.apply(this, options);
+
+        // return early if options is empty or invalid
+        if (!_.isObject(options)) {
+            return;
+        }
+
+        _.mergeDefined(this, {
+            id: options.id,
```
SamvelRaja: `id` should be auto-generated 😨   IMO.

lib/collection/certificate.js
```diff
@@ -0,0 +1,129 @@
...
+        // test the input url against allowed matches
+        return this.matches.test(url);
+    },
+
+    /**
+     * Returns the object representation of the certificate object.
+     *
+     * @return {Object}
+     */
+    toJSON: function() {
```
SamvelRaja: [Property Base](https://github.com/kamalaknn/postman-collection/blob/37ee06cc15ca51e713c00178c87624eb8484f06e/lib/collection/property-base.js#L66) Will handle this.

lib/collection/proxy-config.js
```diff
@@ -102,176 +88,18 @@ _.assign(ProxyConfig.prototype, /** @lends ProxyConfig.prototype */ {
             return;
         }
         _.has(options, 'server') && this.server.update(options.server);
-        _.isString(options.match) && (this.match = options.match);
+        _.isString(options.match) && (this.match = new UrlMatchPattern(options.match));
```
kamalaknn: I thought of the same too, but I did not go down this path, because a `UrlMatchMattern` is a generic class. It takes a pattern and tests input Urls on it. I am not sure, if adding restrictions on what is allowed in the pattern or allowing a feature for it to be customisable is out of the scope of `UrlMatchPattern`.

lib/url-pattern/url-match-pattern-list.js
```diff
@@ -0,0 +1,88 @@
...
+    PropertyList = require('../collection/property-list').PropertyList,
+    Url = require('../collection/url').Url,
+    UrlMatchPattern = require('./url-match-pattern').UrlMatchPattern,
+
+    MATCH_ALL_URLS = UrlMatchPattern.MATCH_ALL_URLS,
+
+    UrlMatchPatternList;
+
+_.inherit((
+  UrlMatchPatternList = function (parent, list) {
```
kamalaknn: Okay will do 👍

lib/url-pattern/url-match-pattern-list.js
```diff
@@ -0,0 +1,88 @@
...
+      UrlMatchPatternList.super_.call(this, UrlMatchPattern, parent, list);
+  }), PropertyList);
+
+_.assign(UrlMatchPatternList.prototype, /** @lends UrlMatchPatternList.prototype */ {
+    toJSON: function() {
```
kamalaknn: Yeah, I mistook `PropertyList.prototype.toJSON` was not calling `member.toJSON()`. Turns out it does.

Removed this now.

lib/url-pattern/url-match-pattern-list.js
kamalaknn: The reason I chose to do this over `.test` is that, it is cheaper to do an equality check on all items. 

Like you said it will have an overhead if the list in hundreds without a *MATCH_ALL_URLS*. But we don't do complex operation for the test. It should be acceptable.

Imagine a long list with *MATCH_ALL_URLS* at the end, then we would have to do slow regex test on all items till we get there.

However, I am not totally poised to this approach, because what is the chance of having a long list with *MATCH_ALL_URLS* at the end. So not sure if we need to optimise for this path.

lib/collection/certificate.js
kamalaknn: It will be, if not passed 😉

lib/collection/proxy-config.js
kamalaknn: Okay 👍 

Looks like `new RegExp()` matches all strings. So `new UrlMatchPattern()` should match all Urls.

lib/url-pattern/url-match-pattern.js
```diff
@@ -0,0 +1,278 @@
...
+     * @returns Match regex object
+     */
+    createMatchPattern: function() {
+        var matchPattern = this.pattern,
+
+            // Check the match pattern of sanity and split it into protocol, host and path
+            match = matchPattern.match(regexes.patternSplit);
+
+        if (!match) {
+        // This ensures it is a invalid match pattern
```
czardoz: minor indentation issue

lib/url-pattern/url-match-pattern.js
```diff
@@ -0,0 +1,278 @@
...
+        return matchRegexObject.protocol === MATCH_ALL ? true : matchRegexObject.protocol === protocol;
+    },
+
+    /**
+     * Tests if the given host string, is allowed by the pattern.
+     *
+     * @param {String=} The host to be checked if the pattern allows.
+     * @return {Boolean=}
+     */
+    testHost: function(host) {
```
czardoz: Some parts of your API accept a `matchRegexObject`, while some don't. Is there any reason for that?

lib/collection/proxy-config.js
kamalaknn: This is redundant. `new UrlMatchPattern()` already defaults to *MATCH_ALL_URLS*.

lib/url-pattern/url-match-pattern-list.js
kamalaknn: Actually, `PropertyList.prototype.toJSON()` does `toJSON` for each property on member, instead of `toJSON` on the member.

This would have been fine for all cases where `member.toJSON()` is hash, but here it is not.

lib/collection/certificate.js
```diff
@@ -0,0 +1,129 @@
...
+* @typedef Certificate~definition
+* @property {String} [name] A name for the certificate
+* @property {Array} [matches] A list of match patterns
+* @property {String} [keyPath] The path on the file system to the file that contains the key
+* @property {String} [pemPath] The path on the file system to the certificate file
+* @property {String} [passphrase] The passphrase for the certificate key
+*
+* @example <caption>JSON definition of an example certificate object</caption>
+* {
+*     name: 'My certificate for example.com',
```
czardoz: How will we store `pfx` type certificates here? + @SamvelRaja

lib/url-pattern/url-match-pattern.js
```diff
@@ -0,0 +1,278 @@
...
+    toString: function() {
+        return this.pattern;
+    },
+
+    /**
+     * Returns a string representation of the match pattern
+     *
+     * @return {String=} pattern
+     */
+    toJSON: function() {
```
SamvelRaja: @shamasis Loop me in while discussing about this API.

lib/collection/certificate.js
```diff
@@ -0,0 +1,129 @@
...
+*     matches: ['https://example.com/*'],
+*     key: { src: '/User/path/to/certificate/key' },
+*     cert: { src: '/User/path/to/certificate' },
+*     passphrase: 'iampassphrase'
```
czardoz: This data structure should also allow us to specify acceptable CA certificates

lib/url-pattern/url-match-pattern-list.js
```diff
@@ -0,0 +1,95 @@
+var _ = require('../util').lodash,
```
czardoz: Is this class really necessary? I'm not sure why we need it at all. It has only one function, `test()`, which IMO can be implemented as a static method on `UrlMatchPattern`. Is there any future use of this? 

@shamasis Thoughts?

lib/url-pattern/url-match-pattern.js
kamalaknn: The reason for taking in `matchRegexObject` as a parameter, is to avoid the regex parsing multiple times within the same method. So that we could use the same `matchRegexObject` for `_testHost`, `_testProtocol` and `_testPath`.

`testHost` is a nice public API, which creates `matchRegexObject` and wraps `_testHost` for anyone else wanting to use it.

lib/url-pattern/url-match-pattern-list.js
kamalaknn: I added this class, so that we can have different approaches to matching a url with a pattern, and a list of patterns.

In case we add [cache mechanisms over internal regex matches](https://github.com/postmanlabs/postman-collection/blob/3706653574b239add3aaed128e8e9c52cfbe2b86/lib/collection/proxy-config.js#L148), this class might be useful over a static method.

lib/url-pattern/url-match-pattern.js
czardoz: :+1:

lib/collection/proxy-config.js
shamasis: My take, matching a subset of protocol can also be thought as "pattern". What do you think?

lib/collection/certificate.js
shamasis: @kamalaknn @SamvelRaja - any reply to this? I don't have much context, but want to ensure all conversations are closed.

lib/collection/certificate.js
shamasis: @kamalaknn @SamvelRaja - any reply to this? I don't have much context, but want to ensure all conversations are closed.

test/unit/certificate.test.js
```diff
@@ -0,0 +1,58 @@
...
+            var rawCert = {
+                    matches: ['https://postman-echo.com/*', 'https://bla.com/*'],
+                    key: { src: '/Users/here' },
+                    cert: { src: '/Users/here' },
+                    passphrase: 'iamhere'
+                },
+                certificate = new Certificate(rawCert),
+                certificateJSON = certificate.toJSON();
+            expect(certificateJSON.matches).to.eql(rawCert.matches);
+            expect(certificate.test('https://bla.com')).to.eql(true);
```
shamasis: are the following four assertions applicable to this test?

lib/url-pattern/url-match-pattern.js
shamasis: Why's to JSON returning string?

lib/url-pattern/url-match-pattern.js
```diff
@@ -0,0 +1,279 @@
...
+/**
+ * UrlMatchPattern allows to create rules to define Urls to match for.
+ * It is based on Google's Match Pattern - https://developer.chrome.com/extensions/match_patterns
+ *
+ * @example <caption>An example UrlMatchPattern</caption>
+ * var matchPattern = new UrlMatchPattern('https://*.google.com/*');
+*/
+_.inherit((
+  UrlMatchPattern = function UrlMatchPattern (pattern) {
+      UrlMatchPattern.super_.apply(this, { pattern: pattern });
```
shamasis: Doing .apply with an object?? Also, why do we need to change the arguments while calling super class. Not a good idea.

The way to handle when you want to allow "string" type arguments is to add `parse` and `unparse` to the instance. See how this is handled in headers - https://github.com/postmanlabs/postman-collection/blob/develop/lib/collection/header.js

lib/url-pattern/url-match-pattern.js
```diff
@@ -0,0 +1,279 @@
...
+
+            // Check the match pattern of sanity and split it into protocol, host and path
+            match = matchPattern.match(regexes.patternSplit);
+
+        if (!match) {
+            // This ensures it is a invalid match pattern
+            return undefined;
+        }
+
+        return ({
```
shamasis: Why this bracket outside objects?

lib/collection/certificate.js
kamalaknn: We analysed possible types of certificates we might want to support. 

Broadly the certificates can be Private Key, and Certificate in a same file (`pfx`) or in individual files (`pem`). So the data structure would look like

```js
{
  key: ...,
  cert: ...
}
```
and 
```
{
    pfx: ...
}
```

I deferred adding support to `pfx` now because `postman-runtime` does not support it now, and it can extended without breaking the current structure.

lib/collection/certificate.js
kamalaknn: @shamasis Same as above, deferred now because `postman-runtime` does not support it now, and can be extended without breaking structure.

lib/url-pattern/url-match-pattern.js
```diff
@@ -0,0 +1,279 @@
...
+     * @private
+     * @param {String} pattern Glob pattern string
+     * @return {RegExp=}
+     */
+    globPatternToRegexp: function(pattern) {
+        // Escape everything except ? and *.
+        pattern = pattern.replace(regexes.escapeMatcher, regexes.escapeMatchReplacement);
+        pattern = pattern.replace(regexes.questionmarkMatcher, regexes.questionmarkReplacment);
+        pattern = pattern.replace(regexes.starMatcher, regexes.starReplacement);
+        return new RegExp(PREFIX_DELIMITER + pattern + POSTFIX_DELIMITER);
```
shamasis: This is slow. Maybe.

try having a function on top with static replacement of matches and a single regex to extract the replacements

test/unit/certificate.test.js
kamalaknn: Not related to this test. Was left here by mistake. Will remove.

lib/collection/certificate.js
```diff
@@ -0,0 +1,114 @@
...
+     * @private
+     */
+    _postman_requiresId: true,
+
+    /**
+     * Tests the url string to check if it matches any of the allowed match patterns.
+     *
+     * @param {String} url The url string for which the proxy match needs to be done.
+     */
+    test: function(url) {
```
shamasis: Test? Test the certificate?

lib/url-pattern/url-match-pattern.js
kamalaknn: In this case (regex with recursive match) a static function with replacements is slower than individual replacements.

https://jsperf.com/string-replace-vs-replace-function

lib/collection/proxy-config.js
czardoz: Yes

lib/collection/certificate.js
```diff
@@ -0,0 +1,161 @@
...
+             * List of match pattern
+             * @type {UrlMatchPatternList}
+             */
+            matches: options.matches && new UrlMatchPatternList({}, options.matches),
+
+            /**
+             * Private Key
+             * @type {{ src: (string) }}
+             */
+            key: options.key,
```
czardoz:
```
key: _.isObject(options.key) ? options.key : { src: options.key },
```

lib/collection/certificate.js
```diff
@@ -0,0 +1,161 @@
...
+
+            /**
+             * Certificate
+             * @type {{ src: (string) }}
+             */
+            cert: options.cert,
```
czardoz: same as above

lib/collection/proxy-config.js
SamvelRaja: 😢

PR #287, opened 2017-03-01
Use member.toJSON in PropertyList.toJSON

kamalaknn:


kamalaknn:
@shamasis, this uses `member.toJSON()` if it exists.

czardoz:
@shamasis Nope, I can't see any, infinite recursion is not possible, because this only changes the PropertyList's `.toJSON()`.

czardoz:
@kamalaknn Also, I've added you to repository collaborators, so you can just create branches here, instead of forking :)

kamalaknn:
@czardoz Okay 👍

PR #288, opened 2017-03-02
Adds proxyConfig to Request.js schema

kamalaknn:
This adds a property `proxyConfig` to `Request.js` so [runtime](https://github.com/postmanlabs/postman-runtime/blob/d9b81ca2f178571f1dc698871535c306ba46a023/lib/requester/request-wrapper.js#L88) can now expose the proxy used for a request, though a public API.

### DIFF
--- a/lib/collection/request.js
+++ b/lib/collection/request.js
@@ -2,6 +2,7 @@ var _ = require('../util').lodash,
     Property = require('./property').Property,
     PropertyList = require('./property-list').PropertyList,
     Url = require('./url').Url,
+    ProxyConfig = require('./proxy-config').ProxyConfig,
     Header = require('./header').Header,
     RequestBody = require('./request-body').RequestBody,
     RequestAuth = require('./request-auth').RequestAuth,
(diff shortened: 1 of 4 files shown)


lib/collection/request.js
```diff
@@ -61,7 +63,12 @@ _.inherit((
             /**
              * @type {RequestAuth}
              */
-            auth: _.createDefined(options, 'auth', RequestAuth)
+            auth: _.createDefined(options, 'auth', RequestAuth),
+
+            /**
+             * @type {ProxyConfig}
+             */
+            proxyConfig: options.proxyConfig && new ProxyConfig(options.proxyConfig)
```
czardoz: just make this `request.proxy`, the `Config` part doesn't help in any way

kamalaknn: Agreed 👍

lib/schema/request.json
```diff
@@ -14,6 +14,9 @@
                 "auth": {
                     "$ref": "#/definitions/auth"
                 },
+                "proxyConfig": {
```
czardoz: Can you also update this to `proxy`? @kamalaknn

kamalaknn: Missed this one. Fixed and pushed.

PR #290, opened 2017-03-03
Feature/prevent meta property serialisation

shamasis:
Refer to commit messages. In short: toJSON of base, no longer publishes meta properties

### DIFF
--- a/lib/collection/property-base.js
+++ b/lib/collection/property-base.js
@@ -1,6 +1,9 @@
 var _ = require('../util').lodash,
     Description = require('./description').Description,
 
+    STRING = 'string',
+    META_PREFIX_CHAR = '_',
+
     PropertyBase; // constructor
 
 /**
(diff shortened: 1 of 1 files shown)


lib/collection/property-base.js
```diff
@@ -97,21 +100,12 @@ _.assign(PropertyBase.prototype, /** @lends PropertyBase.prototype */ {
...
-            // Handle 'PropertyBase's
-            if (value && _.isFunction(value.toJSON)) {
-                accumulator[key] = value.toJSON();
-                return accumulator;
-            }
-
-            // Handle Strings
-            if (_.isString(value)) {
-                accumulator[key] = value;
+            if (PropertyBase.propertyIsMeta(value, key)) {
```
kamalaknn: Can we move this up to avoid plurality check for meta properties?

shamasis: Yes. Right. This should go up.

shamasis:
@kamalaknn let this not get merged yet. Add your toJSON to request. 😞

kamalaknn:
@shamasis Okay.

PR #298, opened 2017-03-15
ensure that cert and key values don't accidentally leak out

czardoz:
Certificate.key.value or Certificate.cert.value should not accidentally be serialized.

### DIFF
--- a/lib/collection/certificate.js
+++ b/lib/collection/certificate.js
@@ -1,5 +1,6 @@
 var _ = require('../util').lodash,
     Property = require('./property').Property,
+    PropertyBase = require('./property-base').PropertyBase,
     Url = require('./url').Url,
     UrlMatchPatternList = require('../url-pattern/url-match-pattern-list').UrlMatchPatternList,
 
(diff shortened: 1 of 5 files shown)


lib/collection/certificate.js
```diff
@@ -135,6 +136,26 @@ _.assign(Certificate.prototype, /** @lends Certificate.prototype */ {
...
+    },
+
+    /**
+     * Allows the serialization of a {@link Certificate}
+     *
+     * This is overridden, in order to ensure that certificate contents are not accidentally serialized,
+     * which can be a security risk.
+     */
+    toJSON: function () {
+        var obj = PropertyBase.toJSON(this);
```
kunagpal: Since we are serialising here anyway, why not:
```javascript
return _.omit(PropertyBase.toJSON(this), ['key.value', 'cert.value']);
```

lib/collection/variable-list.js
```diff
@@ -194,8 +184,8 @@ _.assign(VariableList.prototype, /** @lends VariableList.prototype */ {
         }
 
         // if we request a particular environment, we get the one, else we convert all environments to JSON
-        return arguments.length ? (env && toJSON(env)) : _.map(this._environments, function (environment) {
-            return toJSON(environment);
+        return arguments.length ? (env && PropertyBase.toJSON(env)) : _.map(this._environments, function (environment) {
```
kunagpal: The function in `_.map` can be replaced by `PropertyBase.toJSON.bind(PropertyBase)`.

lib/collection/certificate.js
czardoz: Because it iterates over the JSONified object again, and also goes through the nested properties, which is unnecessary.

kunagpal: Makes sense. `_.unset` could be used to replace the check -> delete sequence here.

PR #315, opened 2017-03-28
Added VariableScope PM API helpers

kunagpal:
This pull request adds the following to `VariableScope.prototype`:

1. `set`: Creates new or updates existing variables in the current context.
2. `get`: Fetches variable values in the current scope by variable name.
3. `unset`: Removes variables from the current scope by variable name.
4. `clear`: Destroys all variables in the current scope. This is a **destructive** action.

> Bonus: Fixed a bug in PropertyList#clear which would cause only the first member to be removed, not all. Associated unit tests are included as well.

### DIFF
--- a/lib/collection/property-list.js
+++ b/lib/collection/property-list.js
@@ -171,9 +171,6 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
 
         !context && (context = this);
 
-        // maintain consistency when being used in bulk from PropertyList#clear
-        predicate = predicate || this.members[0];
-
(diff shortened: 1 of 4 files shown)


lib/collection/property-list.js
```diff
@@ -170,6 +170,10 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
         var match; // to be used if predicate is an ID
 
         !context && (context = this);
+
+        // maintain consistency when being used in bulk from PropertyList#clear
+        predicate = predicate || this.members[0];
```
shamasis: i dont think this is right.

PR #316, opened 2017-03-29
VariableScope#syncVariablesFromScope

kunagpal:


lib/collection/variable-scope.js
```diff
@@ -98,6 +98,20 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
+     * Using this function, one can sync the values of this variable list from an existing VariableScope JSON.
+     *
+     * @param {Object} obj A JSON representation of a VariableScope instance, obtained via a .toJSON call.
+     * @param {Boolean=} [track] A flag that indicates whether or not to track synced variables.
+     *
+     * @returns {Object}
+     */
+    syncVariablesFromScope: function (obj, track) {
+        return this.values.syncFromObject(_.transform(obj.values, function (agg, record) {
+            agg[record.key] = record.value;
```
shamasis: This ignores `type`. You'll need to maintain the types of obj

kunagpal: Has been addressed.

lib/collection/variable-list.js
```diff
@@ -86,25 +86,30 @@ _.assign(VariableList.prototype, /** @lends VariableList.prototype */ {
...
             // we need to create new variable if exists or update existing
             if (list.has(key)) {
-                list.one(key).set(value);
+                list.one(key).update(tmp);
                 ops && ops.updated.push(key);
             }
             else {
-                tmp = { value: value };
                 tmp[indexer] = key;
                 list.add(tmp);
```
shamasis: don't we need to add the key in `tmp`?

lib/collection/variable-scope.js
```diff
@@ -98,6 +98,20 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
+     * Using this function, one can sync the values of this variable list from an existing VariableScope JSON.
+     *
+     * @param {Object} obj A JSON representation of a VariableScope instance, obtained via a .toJSON call.
+     * @param {Boolean=} [track] A flag that indicates whether or not to track synced variables.
+     *
+     * @returns {Object}
+     */
+    syncVariablesFromScope: function (obj, track) {
+        return this.values.syncFromObject(_.transform(obj.values, function (agg, record) {
+            agg[record.key] = { value: record.value, type: record.type };
```
shamasis: disabled variables?

also, since the function says sync from scope, we should check that if it is already a scope instance, we handle that too.

lib/collection/variable-list.js
kunagpal: The added block above takes care of this.

lib/collection/variable-scope.js
kunagpal: Both cases have been handled and tested. 👍

lib/collection/variable-scope.js
```diff
@@ -98,6 +98,25 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
     /**
+     * Using this function, one can sync the values of this variable list from an existing VariableScope JSON.
+     *
+     * @param {Object|VariableScope} obj A VariableScope instance, or it's JSON equivalent.
+     * @param {Boolean=} [track] A flag that indicates whether or not to track synced variables.
+     *
+     * @returns {Object}
+     */
+    syncVariablesFromScope: function (obj, track) {
+        VariableScope.isVariableScope(obj) && (obj = obj.toJSON());
```
shamasis: I would rather love a faster flow. for var scope, I am doing the following in runtime...

- for each variable in source, I am updating it with matching items found in obj, or creating one (while taking care of type)
- for remaining ones, I am deleting from source

PR #317, opened 2017-03-30
Response details standard name

kunagpal:
* The `fromServer` flag in Response details has been replaced with `standardName`.


kunagpal:
@shamasis Looks like NSP is stuck, going ahead with the merge.

PR #323, opened 2017-04-03
Pm api helpers

czardoz:
* Added Response details `standardName` property.
* Added `set`, `get`, `unset` and `clear `helpers to `VariableScope`.
* Fixed a bug in `PropertyList#clear` which caused only the first property of a list to be removed.
* Added `Response#details` helper to retain custom server HTTP reasons.
* Fixed a script resolution bug in `Event#update`.
* Added `Response.isResponse`,  `Request.isRequest` and `Script.isScript` helpers.


lib/collection/variable-scope.js
```diff
@@ -44,6 +44,47 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
+    /**
+     * Creates a new variable, or updates an existing one.
+     *
+     * @param {String} key - The name of the variable to set.
+     * @param {String} value - The value of the variable to be set.
+     */
+    set: function (key, value) {
+        var variable = this.values.one(key);
+
+        // If a variable by the name key was not found in the previous lookup, create a new variable and insert it.
```
czardoz: wrong comment

kunagpal: Oops. 😝

PR #328, opened 2017-04-06
Fixed a bug which caused empty requests to not have required properties

czardoz:


lib/collection/request.js
```diff
@@ -33,19 +33,27 @@ _.inherit((
...
             url: options
         });
 
+        this.update(options);
+    }), Property);
+
+_.assign(Request.prototype, /** @lends Request.prototype */ {
+
+    update: function (options) {
+        var url = _.createDefined(options, 'url', Url) || new Url();
```
shamasis: what's this?

lib/collection/request.js
```diff
@@ -33,19 +33,25 @@ _.inherit((
...
+
+_.assign(Request.prototype, /** @lends Request.prototype */ {
+
+    update: function (options) {
         _.mergeDefined(this, /** @lends Request.prototype */ {
             /**
              * @type {Url}
              */
-            url: _.createDefined(options, 'url', Url),
+            url: _.createDefined(options, 'url', Url, new Url()),
```
czardoz: @shamasis This ensures that if a URL is not defined, we assign an empty one.

shamasis: no. this means, if a url is not defined when sent in update, even the existing url gets overwritten by empty one. :-/

lib/collection/request.js
```diff
@@ -33,19 +33,25 @@ _.inherit((
         // this constructor is intended to inherit and as such the super constructor is required to be executed
         Request.super_.apply(this, arguments);
 
-        // in case definition object is missing, there is no point moving forward
-        if (!options) { return; }
+        // in case definition object is missing, create an empty one
+        options = options || {};
```
shamasis: I also see in other properties that options is not initialised to an object if missing. instead, it is taken care in update function using `_.get`

https://github.com/postmanlabs/postman-collection/blob/66ad51b62dcb2f519b379d027b6864dbab4684f0/lib/collection/header.js

PR #329, opened 2017-04-06
Better static helpers

kunagpal:
All static helpers of the form `XYZ.isXYZ` will now return strict `Boolean` values, even for falsey input.

### DIFF
--- a/test/unit/variable-list.test.js
+++ b/test/unit/variable-list.test.js
@@ -169,9 +169,9 @@ describe('VariableList', function () {
         ]);
 
         it('should work correctly for isVariableList', function () {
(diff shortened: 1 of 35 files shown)


test/unit/variable-list.test.js
```diff
@@ -171,6 +171,7 @@ describe('VariableList', function () {
         it('should work correctly for isVariableList', function () {
             expect(VariableList.isVariableList(variableList)).to.be.ok();
             expect(VariableList.isVariableList({})).to.not.be.ok();
+            expect(VariableList.isVariableList()).to.not.be.ok();
```
czardoz: Check for false here, like in the other test cases

PR #332, opened 2017-04-14
Added an intermediate base class for auths to inherit from.

czardoz:
- `RequestAuthBase` inherits from `PropertyBase`
- All auth definitions inherit from `RequestAuthBase` (through a dynamic intermediate class)
- XYZAuth has the parameters as direct properties. E.g:

        request.auth.basic.username === 'username' // true
        request.auth.basic.password === 'password' // true
- `Item` now has a new method
    - `getAuth()`: Returns an instance of the currently selected authtype (searches parent folders as well)
- `PropertyBase` has two new methods
    - `.lookupOwner('prop-to-find')`: Finds the closest ancestor which has the given property
    - `.lookup('prop-to-lookup')`: Returns the value of the given property from the closest ancestor
- `RequestAuth` has a new method,
    - `.current()` returns an instance of the currently selected auth.


lib/collection/request-auth-base.js
```diff
@@ -0,0 +1,126 @@
...
+
+_.extend(RequestAuthBase.prototype, /** @lends RequestAuthState.prototype */ {
+
+    /**
+     * Sets a value in the state.
+     *
+     * @param name
+     * @param value
+     */
+    set: function (name, value) {
```
czardoz: The problem with this is, if someone wants to update the auth parameters itself, they have to do:
```
request.auth.basic.username = 'new-username' // this will be serialized with the collection
```
whereas 
```
request.auth.basic.set('username', 'new-username')  // this will be set in state, so not serialized.
```

Seems rather counter-intuitive in terms of usage.

shamasis: I agree. This should not be the case. Let me re rethink.

lib/collection/item.js
```diff
@@ -134,12 +134,19 @@ _.assign(Item.prototype, /** @lends Item.prototype */ {
...
-    processAuth: function () {
-        return this.request.authorize();
+    getAuth: function () {
+        var auth;
+
+        if (this.request.auth && (auth = this.request.auth.current())) {
+            return auth;
+        }
+
+        auth = this.lookup('auth');
```
shamasis: find in parent maybe???

lib/collection/property-base.js
```diff
@@ -82,6 +82,42 @@ _.assign(PropertyBase.prototype, /** @lends PropertyBase.prototype */ {
     },
 
     /**
+     * Tries to find the given property locally, and then proceeds to lookup in each parent,
+     * going up the chain as necessary.
+     *
+     * @param {String} property
+     */
+    lookup: function (property) {
+        var owner = this.lookupOwner(property);
```
shamasis: find parent containing?

PR #335, opened 2017-04-19
addition of header list property

shamasis:
The goal was to have dedicated header list object so that we can **begin** to remove header specific helpers from response/request and make them common across both.

- Changes are done to PropertyList that was found to be generic and easy to work with on all lists while making HeaderList.
- Added concepts of valueOf in Properties and made lists aware of it so that getters can be easily written.
- Changes were made in places where they depended on Header.XYZ static functions
- Fixed the behaviour of computing header size by adhering to proper segregation of concerns.
- Header.headerValue is doscontinued in favour of HeaderList.get and Header (instance) .valueOf

### DIFF
--- a/lib/collection/header-list.js
+++ b/lib/collection/header-list.js
@@ -1,0 +1,55 @@
+var _ = require('../util').lodash,
+    PropertyList = require('./property-list').PropertyList,
+    Header = require('./header').Header,
+
+    HeaderList;
+
+_.inherit((
+
+  /**
+   * @constructor
+   * @extends {PropertyList}
+   */
+  HeaderList = function (parent, headers) {
+      // this constructor is intended to inherit and as such the super constructor is required to be executed
+      HeaderList.super_.call(this, Header, parent, headers);
+  }), PropertyList);
+
+_.assign(HeaderList.prototype, /** @lends HeaderList.prototype */ {
+    /**
+     * Gets size of a list of headers excluding standard header prefix.
+     *
+     * @returns {Number}
+     */
+    contentSize: function () {
+        var raw = '';
+        raw += Header.unparse(this, '\r\n');
+        raw += '\r\n\r\n';
+        return raw.length;
+    }
+});
+
+_.assign(HeaderList, /** @lends HeaderList */ {
+    /**
+     * Defines the name of this property for internal use.
+     * @private
+     * @readOnly
+     * @type {String}
+     */
+    _postman_propertyName: 'HeaderList',
+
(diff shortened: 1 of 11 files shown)


lib/collection/header.js
```diff
@@ -182,12 +195,20 @@ _.assign(Header, /** @lends Header */ {
...
     unparse: function (headers, separator) {
         if (!_.isArray(headers) && !PropertyList.isPropertyList(headers)) {
-            return '';
+            return E;
         }
 
-        return headers.map(function (header) {
-            return header.key + ': ' + header.value;
-        }).join(separator ? separator : '\n');
+        return headers.map(Header.unparseSingle).join(separator ? separator : '\n');
```
kunagpal: `separator || '\n'`

lib/collection/header.js
```diff
@@ -182,12 +195,20 @@ _.assign(Header, /** @lends Header */ {
...
+    },
+
+    /**
+     * Unparses a single Header.
+     * @param {String} header
+     * @returns {String}
+     */
+    unparseSingle: function (header) {
+        if (!_.isObject(header)) { return E; }
```
kunagpal: The flow of the if condition can be reversed so that `return E;` is the fallback instead of the first check.

lib/collection/property-list.js
```diff
@@ -379,6 +393,50 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
+                if (!member.hasOwnProperty(key)) { return; }
+                if (excludeDisabled && member.disabled) { return; }
+
+                obj[member[key]] = member.valueOf();
+            });
+        }
+        else {
+            _.forOwn(this.reference, function (member, prop) {
+                _.isArray(member) && (member = _.last(member));
+                if (excludeDisabled && member.disabled) { // do no process disabled objects
```
kunagpal: typo in comment

lib/collection/header-list.js
```diff
@@ -0,0 +1,57 @@
...
+  }), PropertyList);
+
+_.assign(HeaderList.prototype, /** @lends HeaderList.prototype */ {
+    /**
+     * Gets size of a list of headers excluding standard header prefix.
+     *
+     * @returns {Number}
+     */
+    contentSize: function () {
+        var raw = '';
```
kunagpal: `raw` can directly be assigned with `Header.unparse(this, '\r\n')`. Also, `\r\n` can be assigned to `CRLF` and reused.

PR #343, opened 2017-04-26
PropertyBase#setParent

kunagpal:


lib/collection/property-base.js
```diff
@@ -167,6 +169,15 @@ _.assign(PropertyBase.prototype, /** @lends PropertyBase.prototype */ {
...
         return this && this.__parent && this.__parent.__parent || undefined;
+    },
+
+    /**
+     * Accepts an object and sets it as the parent of the current property.
+     *
+     * @param {Object} parent The object to set as parent.
+     */
+    setParent: function (parent) {
+        _.assignLocked(this, __PARENT, parent);
```
czardoz: Only do this if no parent currently exists, because if one is already there, this will throw an error.

We're assigning the property with `writable: false`

test/unit/property-base.test.js
```diff
@@ -143,4 +143,20 @@ describe('PropertyBase', function () {
...
     });
+
+    describe('.setParent()', function () {
+        it('should set a provided parent correctly', function () {
+            var parent = { foo: 'bar' },
+                base = new sdk.PropertyBase();
+
+            base.setParent(parent);
+
+            expect(Object.getOwnPropertyDescriptor(base, '__parent')).to.eql({
```
czardoz: why can't you just access it using `base.__parent`?

kunagpal: The test also checks for `writable: false`.

lib/collection/property-base.js
kunagpal: Has been altered to use `_.assignHidden` instead.

PR #351, opened 2017-05-03
drop inheritence support from varlist

shamasis:
This was a private feature and not used at all. This was prior to the days of `VariableScope`. I am removing the feature completely and removed relevant tests.

- some tests were wrongly written to use inheritance for normal tests. fixed those
- bug fix: certificate constructor was not being called with proper arguments @SamvelRaja 
- proxy test spec file name was wrong.


czardoz:
@kunagpal Merge?

PR #352, opened 2017-05-05
Header inherits from Property instead of PropertyBase

kunagpal:


czardoz:
@kunagpal Can you update the changelog and add an unreleased section?

PR #355, opened 2017-05-05
Added Item#setRequest, Item#setResponses

kunagpal:


lib/collection/item.js
```diff
@@ -134,6 +134,24 @@ _.assign(Item.prototype, /** @lends Item.prototype */ {
...
+    setRequest: function (request) {
+        this.request = new Request(request);
+    },
+
+    /**
+     * Sets the list of responses for the current item to the provided responses array.
+     *.
+     * @param {Object[]} responses - The responses Array to be set within the current item.
+     */
+    setResponses: function (responses) {
```
shamasis: This will always reset the list.

lib/collection/item.js
```diff
@@ -134,6 +134,24 @@ _.assign(Item.prototype, /** @lends Item.prototype */ {
     _postman_requiresId: true,
 
     /**
+     * Sets the request for the current item to the provided request object.
+     *
+     * @param {Object} request - The request object to be set within the current item.
+     */
+    setRequest: function (request) {
+        this.request = new Request(request);
+    },
```
shamasis: What if an SDK request needs to be added?

PR #358, opened 2017-05-08
Add ability to search multiple layers for variable resolution

vjrngn:
1. Add private method to include additional layers to variable scope
2. Modify `VariableScope.get` to look locally and across parents.

### DIFF
--- a/lib/collection/variable-scope.js
+++ b/lib/collection/variable-scope.js
@@ -74,14 +74,6 @@ _.inherit((
         this.values = new VariableList(this, VariableList.isVariableList(values) ? values.toJSON() : values);
         // in above line, we clone the values if it is already a list. there is no point directly using the instance of
         // a variable list since one cannot be created with a parent reference to begin with.
-
-        _.assign(this, /** @lends VariableScope.prototype */ {
-            /**
-             * @private
-             * @type {Array}
-             */
-            layers: []
-        });
     }), Property);
 
 _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
@@ -140,18 +132,16 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
      * @returns {*} The value of the specified variable across scopes.
      */
     get: function (key) {
-        var variable = this.values.one(key);
+        var variable = this.values.one(key),
+            i,
+            ii;
 
-        // if a variable does not exist in local scope, we search all the layers and find the first occurence.
-        if (!variable) {
-            _.forEach(this.layers, function (layer) {
-                variable = layer.one(key);
-
-                // stop at the first occurence
-                if (variable && variable.key === key) {
-                    return false;
-                }
-            });
+        // if a variable does not exist in local scope, we search all the layers and return the first occurence.
+        if (!variable && this._layers) {
+            for (i = 0, ii = this._layers.length; i < ii; i++) {
+                variable = this._layers[i].one(key);
+                if (variable) { break; }
+            }
         }
 
         return variable ? variable.valueOf() : undefined;
@@ -234,26 +224,29 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
             delete obj.value;
         }
 
+        // ensure that the concept of layers is not exported as JSON. JSON cannot retain references and this will end up
+        // being a pointless object post JSONification.
+        if (obj._layers) {
+            delete obj._layers;
+        }
+
         return obj;
     },
 
     /**
      * Adds a variable list to the current instance in order to increase the surface area of variable resolution.
      * This enables consumers to search across scopes (eg. environment and globals).
+     * @private
      *
-     * @param {VariableList|Array} [definition]
+     * @param {VariableList} [list]
      */
-    _addLayer: function (definition) {
-        var values;
-
-        if (_.isArray(definition) || VariableList.isVariableList(definition)) {
-            definition = { values: definition };
+    addLayer: function (list) {
+        if (!VariableList.isVariableList(list)) {
+            return;
         }
 
-        values = definition && definition.values;
-        this.layers.push(
-            new VariableList(this, VariableList.isVariableList(values) ? values.toJSON() : values)
-        );
+        !this._layers && (this._layers = []);
+        this._layers.push(list);
     }
 });
 
--- a/lib/collection/variable-scope.js
+++ b/lib/collection/variable-scope.js
@@ -74,6 +74,14 @@ _.inherit((
         this.values = new VariableList(this, VariableList.isVariableList(values) ? values.toJSON() : values);
         // in above line, we clone the values if it is already a list. there is no point directly using the instance of
         // a variable list since one cannot be created with a parent reference to begin with.
+
+        _.assign(this, /** @lends VariableScope.prototype */ {
+            /**
+             * @private
+             * @type {Array}
+             */
+            layers: []
+        });
     }), Property);
 
 _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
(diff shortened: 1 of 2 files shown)


lib/collection/variable-scope.js
```diff
@@ -214,6 +235,25 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
+     * @param {VariableList|Array} [definition]
+     */
+    _addLayer: function (definition) {
+        var values;
+
+        if (_.isArray(definition) || VariableList.isVariableList(definition)) {
+            definition = { values: definition };
+        }
+
+        values = definition && definition.values;
```
kunagpal: The `var` statement on line 247 can be shifted here.

lib/collection/variable-scope.js
```diff
@@ -74,6 +74,14 @@ _.inherit((
         this.values = new VariableList(this, VariableList.isVariableList(values) ? values.toJSON() : values);
         // in above line, we clone the values if it is already a list. there is no point directly using the instance of
         // a variable list since one cannot be created with a parent reference to begin with.
+
+        _.assign(this, /** @lends VariableScope.prototype */ {
```
shamasis: single assignment. avoid lodash until more than a few are needed.

also - this is private and very low use-case as such an extra array does not need to be created for every variable scope. we can create it on first `.addLayer` call

test/unit/variable-scope.test.js
```diff
@@ -422,4 +422,79 @@ describe('VariableScope', function () {
...
+                key: 'var-2-layerTwo',
+                value: 'var-2-layerTwo-value'
+            }]);
+
+        it('adds a variable list to the current scope', function () {
+            var scope = new VariableScope(layerOne);
+            scope._addLayer(layerTwo);
+
+            expect(scope.layers.length).to.be(1);
+            expect(VariableList.isVariableList(scope.layers[0])).to.be.ok();
```
kunagpal: This test should be enhanced to actually check for values from the first layer.

lib/collection/variable-scope.js
```diff
@@ -126,13 +134,26 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
-     * @returns {*} The value of the specified variable in the current context.
+     * @param {String} key - The name of the variable to get.
+     * @returns {*} The value of the specified variable across scopes.
      */
     get: function (key) {
         var variable = this.values.one(key);
+
+        // if a variable does not exist in local scope, we search all the layers and find the first occurence.
+        if (!variable) {
+            _.forEach(this.layers, function (layer) {
```
czardoz: Better use a `for` loop here, this is a low-level function, so we don't need the additional overhead of multiple function calls.

Also, then you can simply return from the loop itself, making the code simpler.

lib/collection/variable-scope.js
```diff
@@ -126,13 +134,26 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
+                variable = layer.one(key);
+
+                // stop at the first occurence
+                if (variable && variable.key === key) {
```
shamasis: why `variable.key === key`?

1. `layer.one(key)` already returns based on key
2. if underlying list as case-insensitivity configured, this might break.

lib/collection/variable-scope.js
```diff
@@ -214,6 +235,25 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
         return obj;
+    },
+
+    /**
+     * Adds a variable list to the current instance in order to increase the surface area of variable resolution.
+     * This enables consumers to search across scopes (eg. environment and globals).
+     *
+     * @param {VariableList|Array} [definition]
+     */
+    _addLayer: function (definition) {
```
shamasis: What `_addLayer`?

lib/collection/variable-scope.js
```diff
@@ -74,6 +74,14 @@ _.inherit((
         this.values = new VariableList(this, VariableList.isVariableList(values) ? values.toJSON() : values);
         // in above line, we clone the values if it is already a list. there is no point directly using the instance of
         // a variable list since one cannot be created with a parent reference to begin with.
+
+        _.assign(this, /** @lends VariableScope.prototype */ {
+            /**
+             * @private
+             * @type {Array}
+             */
+            layers: []
```
czardoz: Should this be a `PropertyList` of type `VariableList`?

lib/collection/variable-scope.js
```diff
@@ -214,6 +235,25 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
         }
 
         return obj;
+    },
+
+    /**
```
shamasis: This is a `@private` function

lib/collection/variable-scope.js
```diff
@@ -214,6 +235,25 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
+     * Adds a variable list to the current instance in order to increase the surface area of variable resolution.
+     * This enables consumers to search across scopes (eg. environment and globals).
```
shamasis: "users", not "customers"

lib/collection/variable-scope.js
```diff
@@ -214,6 +235,25 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
+     *
+     * @param {VariableList|Array} [definition]
+     */
+    _addLayer: function (definition) {
+        var values;
+
+        if (_.isArray(definition) || VariableList.isVariableList(definition)) {
```
shamasis: always accept a variableList, no need to worry about non-sdk properties.

lib/collection/variable-scope.js
```diff
@@ -214,6 +235,25 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
+            definition = { values: definition };
+        }
+
+        values = definition && definition.values;
+        this.layers.push(
```
shamasis: this needs to be held by reference. so that if original scopes change, these change too.

lib/collection/variable-scope.js
vjrngn: Just an Array<VariableList>

test/unit/variable-scope.test.js
vjrngn: I don't think that's required as the two `lists` are being store separately. The one initialized in the constructor is `this.values` and the layers are added as a private property as `this._layers`

lib/collection/variable-scope.js
```diff
@@ -56,16 +57,22 @@ _.inherit((
...
-    VariableScope = function PostmanVariableScope (definition) {
+    VariableScope = function PostmanVariableScope (definition, layers) {
         // in case the definition is an array (legacy format) or existing as list, we convert to actual format
         if (_.isArray(definition) || VariableList.isVariableList(definition)) {
             definition = { values: definition };
         }
 
+        // we accept parent scopes to increase search area. Here we normalize the argument to be an array
+        // so we can easily loop though them and add them to the instance.
+        layers && !_.isArray(layers) && (layers = [layers]);
```
kunagpal: This could be compressed into:
```js
layers = _.castArray(layers);
```

lib/collection/variable-scope.js
```diff
@@ -74,6 +81,16 @@ _.inherit((
         this.values = new VariableList(this, VariableList.isVariableList(values) ? values.toJSON() : values);
         // in above line, we clone the values if it is already a list. there is no point directly using the instance of
         // a variable list since one cannot be created with a parent reference to begin with.
+
+        this._layers = [];
+
+        if (layers) {
+            for (i = 0, ii = layers.length; i < ii; i++) {
```
kunagpal: Could be written as:
```js
layers && for...
```

lib/collection/variable-scope.js
```diff
@@ -74,6 +81,16 @@ _.inherit((
...
+                if (VariableList.isVariableList(layers[i])) {
```
kunagpal: Could be rewritten as:
```js
VariableList.isVariableList(layers[i]) && this._layers.push(layers[i]);
```

lib/collection/variable-scope.js
vjrngn: I agree. Just thought this looked cleaner.

PR #362, opened 2017-05-10
Add convenience method to variable scope to convert variable list to obj

vjrngn:
Add `variableScope.toObject` as a proxy to `propertyList.toObject`.  This is done to accommodate the new `pmapi`.

Why ?
The new pmapi exposes pm.variables as an instance of VariableScope. Having `pm.variables.variables` is redundant and does not make a good API. `VariableScope.variables` is deprecated as of version `1.2.5.*`

### DIFF
--- a/test/unit/variable-scope.test.js
+++ b/test/unit/variable-scope.test.js
@@ -562,5 +562,18 @@ describe('VariableScope', function () {
                 'key3': 'val3'
             });
         });
+
+        it('uses the last found key-val pair should a duplicate key exists', function () {
+            var scope = new VariableScope(keyVals.concat({
+                key: 'key3',
+                value: 'duplicate-val3'
+            }));
+
+            expect(scope.toObject()).to.eql({
+                'key1': 'val1',
+                'key2': 'val2',
+                'key3': 'duplicate-val3'
+            });
+        });
     });
 });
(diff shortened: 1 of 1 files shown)


lib/collection/variable-scope.js
```diff
@@ -131,6 +132,16 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
     },
 
     /**
+     * Converts a list of Variables into an object where key is `_postman_propertyIndexKey` and value is determined
+     * by the `valueOf` function
+     *
+     * @return {Object}
+     */
+    toObject: function (excludeDisabled, caseSensitive) {
```
shamasis: Good that we kept the parameters `excludeDisabled, caseSensitive` private. No need to document these though.

lib/collection/variable-scope.js
```diff
@@ -131,6 +132,16 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
+        return this.values.toObject(excludeDisabled, caseSensitive);
```
czardoz: Is this equivalent to `this.values.syncToObject({})`?

kunagpal: Not exactly, this returns an object of the form:
```json
{
   "0": "...",
   "1": "..."
}
```

PR #372, opened 2017-05-29
enhanced the .has() function in PropertyList, to also allow checking …

czardoz:
…for values (if provided)


lib/collection/property-list.js
```diff
@@ -360,12 +360,47 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
      * Check whether an item exists in this list
      *
      * @param {String|PropertyList.Type} item
+     * @param value
```
kunagpal: JSDoc

lib/collection/property-list.js
```diff
@@ -360,12 +360,47 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
+
+        // if no value is provided, just check if item exists
+        if (arguments.length === 1) {
+            return Boolean(_.isArray(match) ? match.length : match);
+        }
+
+        // If this property allows multiple values and we get an array, we need to iterate through it and see
+        // if any element matches.
+        if (this._postman_propertyAllowsMultipleValues && _.isArray(match)) {
+            for (i = 0; i < match.length; i++) {
```
kunagpal: Can be replaced with `_.forEach`.

czardoz: For loop is faster, and this is a lower level function, so we'll not introduce extra function calls.

kunagpal: Alright.

PR #373, opened 2017-05-29
Fixed falsy path variable resolution

kunagpal:
With this change, non-string path variable values will not be considered for resolution.

### DIFF
--- a/lib/collection/url.js
+++ b/lib/collection/url.js
@@ -245,7 +245,7 @@ _.assign(Url.prototype, /** @lends Url.prototype */ {
                 if (_.startsWith(segment, PATH_VARIABLE_IDENTIFIER)) {
                     value = self.variables.one(segment.slice(1));
                 }
-                _.isString(value && value.value) && res.push(value);
+                res.push(_.isString(value && value.value) ? value : segment);
             }, []),
             path = segmentArray.join(PATH_SEPARATOR);
         return _.startsWith(path, PATH_SEPARATOR) ? path : PATH_SEPARATOR + path;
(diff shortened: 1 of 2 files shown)


test/unit/url.test.js
```diff
@@ -392,6 +392,25 @@ describe('Url', function () {
...
+                    { id: 'alpha', value: 'get' },
+                    { id: 'beta', value: null },
+                    { id: 'gamma', value: NaN },
+                    { id: 'gamma', value: undefined },
+                    { id: 'epsilon', value: [] },
+                    { id: 'phi', value: {} }
+                ]
+            });
+
+            expect(url.getPath()).to.eql('/get');
```
czardoz: The behavior should result in the path `/get/:beta/:gamma/:delta/:epsilon/:phi`

kunagpal: Okay.

lib/collection/url.js
```diff
@@ -240,13 +240,13 @@ _.assign(Url.prototype, /** @lends Url.prototype */ {
...
         var self = this,
-            segmentArray = (options && options.unresolved) ? this.path : _.map(this.path, function (segment) {
+            segmentArray = (options && options.unresolved) ? this.path : _.transform(this.path, function (res, segment) {
                 var value;
                 if (_.startsWith(segment, PATH_VARIABLE_IDENTIFIER)) {
                     value = self.variables.one(segment.slice(1));
                 }
-                return value ? value : segment;
-            }),
+                res.push(_.isString(value && value.value) ? value : segment);
```
czardoz:
```
_.isString(value.valueOf()) ? value.valueOf() : segment);
```

I would suggest renaming the variable `value` to `variable`.

kunagpal: Has been fixed.

PR #374, opened 2017-05-29
Added support for NTLM auth

czardoz:


test/unit/request-auth.test.js
```diff
@@ -106,6 +106,28 @@ describe('RequestAuth', function () {
...
+                    auth: {
+                        type: 'ntlm',
+                        ntlm: {
+                            username: 'testuser',
+                            password: 'testpass',
+                            domain: 'testdomain',
+                            workstation: 'sample.work'
+                        }
+                    },
+                    url: 'httpbin.org/get'
```
kunagpal: Although this doesn't actually result in a request, it should point to `echo/get`

lib/collection/request-auth/ntlm.js
```diff
@@ -0,0 +1,26 @@
...
+            workstation: options.workstation
+        });
+    },
+
+    /**
+     * Does nothing to the request, exists because we want to keep the API consistent.
+     *
+     * @note NTLM requests need intermediate signing information from the server, so we do not make any changes to
+     * the request here.
+     * @param request
```
kunagpal: JSDoc

PR #384, opened 2017-06-16
`variableScope.toObject` returns data from all layers

vjrngn:


lib/collection/variable-scope.js
```diff
@@ -138,7 +138,15 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
      * @return {Object}
      */
     toObject: function (excludeDisabled, caseSensitive) {
-        return this.values.toObject(excludeDisabled, caseSensitive);
+        var mergedLayers = {},
+            length = this._layers.length,
```
shamasis: presence of `_layers` is guaranteed?

czardoz: is guaranteed by the constructor

shamasis:
@kunagpal fix whatever is failing and merge.

PR #390, opened 2017-06-23
update proxyConfig structure

harryi3t:
Changes made:
- removed `server` and added `host`, `port` & `protocols` properties
- any protocol will be stripped away from both the match pattern as well the host value passed to the constructor 
  #### example
  ```js
    {
        // match pattern auto conversion
         'http://example.com/*'. : 'example.com/*',
         'example.com/*'.        : 'example.com/*',
        
        // host auto conversion
        'http://proxy.com'.      : 'proxy.com',
        'foo://proxy.com'.       : 'proxy.com'
    }
  ```
   **PS:** While stripping away protocol from `host` or `match` pattern, proxyConfig does not update it's `protocols` property.

Default values
-  protocol: `http`
-  port:  8080
- match: {pattern: '<all_urls>'}  // which is equivalent to *

### DIFF
--- a/lib/collection/proxy-config-list.js
+++ b/lib/collection/proxy-config-list.js
@@ -16,7 +16,7 @@ _.inherit((
      * @example <caption>Create a new ProxyConfigList</caption>
      * var ProxyConfigList = require('postman-collection').ProxyConfigList,
      *     myProxyConfig = new ProxyConfigList({}, [
-     *              {match: 'example.com/*',host: 'proxy.com', port: 8080, protocols: ['https], tunnel: true},
+     *              {match: 'example.com/*',host: 'proxy.com', port: 8080, protocols: ['https'], tunnel: true},
      *              {match: 'example2.com/*',host: 'proxy2.com'},
      *          ]);
      */
--- a/lib/collection/proxy-config-list.js
+++ b/lib/collection/proxy-config-list.js
@@ -16,8 +16,8 @@ _.inherit((
      * @example <caption>Create a new ProxyConfigList</caption>
      * var ProxyConfigList = require('postman-collection').ProxyConfigList,
      *     myProxyConfig = new ProxyConfigList({}, [
-     *              {match: 'example.com/*',host: 'proxy.com', port: 8080, protocols: ['https'], tunnel: true},
-     *              {match: 'example2.com/*',host: 'proxy2.com'},
+     *              {match: 'example.com/*', host: 'proxy.com', port: 8080, protocols: ['https'], tunnel: true},
+     *              {match: 'example2.com/*', host: 'proxy2.com'},
      *          ]);
      */
     ProxyConfigList = function PostmanProxyConfigList (parent, populate) {
--- a/lib/collection/proxy-config-list.js
+++ b/lib/collection/proxy-config-list.js
@@ -16,8 +16,8 @@ _.inherit((
      * @example <caption>Create a new ProxyConfigList</caption>
      * var ProxyConfigList = require('postman-collection').ProxyConfigList,
      *     myProxyConfig = new ProxyConfigList({}, [
-     *              {match: 'https://example.com/*',server: 'https://proxy.com',tunnel: true},
-     *              {match: 'https://example2.com/*',server: 'http://proxy2.com'},
+     *              {match: 'example.com/*',host: 'proxy.com', port: 8080, protocols: ['https], tunnel: true},
+     *              {match: 'example2.com/*',host: 'proxy2.com'},
      *          ]);
      */
     ProxyConfigList = function PostmanProxyConfigList (parent, populate) {
@@ -30,7 +30,7 @@ _.assign(ProxyConfigList.prototype, /** @lends ProxyConfigList.prototype */ {
     /**
      * Matches and gets the proxy config for the particular url
      * @returns {ProxyConfig~definition=} The matched proxyConfig object
-     * @param {String=} [url] The url for which the proxy config needs to be fetched
+     * @param {URL=} [url] The url for which the proxy config needs to be fetched
      */
     resolve: function(url) {
         // url must be either string or an instance of url.
--- a/lib/collection/proxy-config.js
+++ b/lib/collection/proxy-config.js
@@ -63,7 +63,7 @@ _.inherit((
         ProxyConfig.super_.call(this, options);
 
         // Assign defaults before proceeding
-        _.assign(this, /** @lends ProxyConfig.prototype */ {
+        _.assign(this, /** @lends ProxyConfig */ {
             /**
              * The url mach for which the proxy has been associated with.
              * @type {String}
--- a/lib/collection/proxy-config.js
+++ b/lib/collection/proxy-config.js
@@ -1,7 +1,7 @@
 var _ = require('../util').lodash,
     Property = require('./property').Property,
     UrlMatchPattern = require('../url-pattern/url-match-pattern').UrlMatchPattern,
-    MATCH_ALL_URLS = '<all_urls>',
+    MATCH_ALL_URLS = UrlMatchPattern.MATCH_ALL_URLS,
     DEFAULT_PROTOCOLS = ['http'],
     regexes = {
         protocolMatcher: /.*:\/\//
--- a/lib/collection/proxy-config.js
+++ b/lib/collection/proxy-config.js
@@ -161,14 +161,6 @@ _.assign(ProxyConfig, /** @lends ProxyConfig */ {
     _postman_propertyName: 'ProxyConfig',
 
     /**
-     * Specify the key to be used while indexing this object
-     * @private
-     * @readOnly
-     * @type {String}
-     */
-    _postman_propertyIndexKey: 'match',
-
-    /**
      * Check whether an object is an instance of PostmanItem.
      *
      * @param {*} obj
--- a/lib/collection/proxy-config.js
+++ b/lib/collection/proxy-config.js
@@ -1,8 +1,11 @@
 var _ = require('../util').lodash,
     Property = require('./property').Property,
-    Url = require('./url').Url,
     UrlMatchPattern = require('../url-pattern/url-match-pattern').UrlMatchPattern,
-
+    MATCH_ALL_URLS = '<all_urls>',
+    DEFAULT_PROTOCOLS = ['http'],
+    regexes = {
+        protocolMatcher: /.*:\/\//
+    },
     ProxyConfig;
 
 
@@ -12,15 +15,19 @@ var _ = require('../util').lodash,
 * Proxy instance.
 * @typedef ProxyConfig~definition
 *
-* @property {String=} [url = *] The match for which the proxy needs to be configured. `*` represents select all.
-* @property {Url=} [server = new Url()] The proxy server url.
+* @property {String=} [match = '<all_url>'] The match for which the proxy needs to be configured.
+* @property {String=} [host = ''] The proxy server url.
+* @property {Integer=} [port = 8080] The proxy server port number.
+* @property {Array=} [protocols = ['http']] The protocols that the proxy server supports
 * @property {Boolean=} [tunnel = false] The tunneling option for the proxy request.
 * @property {Boolean=} [disabled = false] To override the proxy for the particiular url, you need to provide true.
 *
 * @example <caption>JSON definition of an example proxy object</caption>
 * {
-*     match: 'https://example.com/*',
-*     server: 'https://proxy.com',
+*     match: 'example.com/*',
+*     host: 'proxy.com',
+*     port: 8080
+*     protocols: ['https']
 *     tunnel: true,
 *     disabled: false
 * }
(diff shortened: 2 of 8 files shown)


harryi3t:
# Updated proxyConfig

image

# Updated proxyConfigList
image

lib/collection/proxy-config.js
```diff
@@ -1,8 +1,11 @@
 var _ = require('../util').lodash,
     Property = require('./property').Property,
-    Url = require('./url').Url,
     UrlMatchPattern = require('../url-pattern/url-match-pattern').UrlMatchPattern,
-
+    MATCH_ALL_URLS = '<all_urls>',
```
SamvelRaja: This should be taken from UrlMatchPattern

lib/collection/proxy-config.js
```diff
@@ -85,9 +106,22 @@ _.assign(ProxyConfig.prototype, /** @lends ProxyConfig.prototype */ {
...
+        _.isInteger(options.port) && (options.port >= 0 && options.port <= 65535) && (this.port = options.port);
+        _.isBoolean(options.tunnel) && (this.tunnel = options.tunnel);
+
+        // match pattern
+        if (_.isString(options.match)) {
+            if (_.isEmpty(options.match) || options.match === MATCH_ALL_URLS) {
+                this.match = new UrlMatchPattern();
+            }
+            else {
+                this.match = new UrlMatchPattern('://' + options.match.replace(regexes.protocolMatcher, ''));
```
SamvelRaja: Shall we have the '://' handling in the UrlMatchPattern itself? which behaves differently, if the protocols are provided seperately?

SamvelRaja: Also cache the host and use it here, we are doing the .replace twice now.

czardoz:
Is this a breaking change?

lib/collection/proxy-config.js
```diff
@@ -60,10 +68,22 @@ _.inherit((
...
+             * @type {String}
+             */
+            host: '',
+
+            /**
+             * The proxy server port number
+             * @type {Integer}
              */
-            server: new Url(),
+            port: 8080,
```
SamvelRaja: Make it as a Constant.

lib/collection/proxy-config.js
```diff
@@ -85,9 +105,21 @@ _.assign(ProxyConfig.prototype, /** @lends ProxyConfig.prototype */ {
         if (!_.isObject(options)) {
             return;
         }
-        _.has(options, 'server') && this.server.update(options.server);
-        _.isString(options.match) && (this.match = new UrlMatchPattern(options.match));
-        _.has(options, 'server') && (this.tunnel = _.isBoolean(options.tunnel) ? options.tunnel : false);
+        _.isString(options.host) && (this.host = options.host.replace(regexes.protocolMatcher, ''));
+        _.isInteger(options.port) && (options.port >= 0 && options.port <= 65535) && (this.port = options.port);
```
SamvelRaja: `65535`  Constant Pls :)

lib/collection/proxy-config.js
```diff
@@ -1,8 +1,13 @@
 var _ = require('../util').lodash,
     Property = require('./property').Property,
-    Url = require('./url').Url,
     UrlMatchPattern = require('../url-pattern/url-match-pattern').UrlMatchPattern,
-
+    DEFAULT_PROTOCOLS = ['http'],
+    DEFAULT_PORT = 8080,
```
czardoz: Why do this for integers?

The V8 Engine is smart enough to inline the integers to machine code when it sees them. By making variables, we're adding an extra lookup.

It's not a big deal, but if we don't need it, might as well not do it.

lib/collection/proxy-config.js
```diff
@@ -52,6 +63,15 @@ _.inherit((
         ProxyConfig.super_.call(this, options);
 
         // Assign defaults before proceeding
+        _.assign(ProxyConfig.prototype, /** @lends ProxyConfig.prototype */ {
```
czardoz: Why are you doing this inside every constructor call? You need to do it just once, outside the constructor block.

lib/collection/proxy-config.js
```diff
@@ -85,9 +117,23 @@ _.assign(ProxyConfig.prototype, /** @lends ProxyConfig.prototype */ {
         if (!_.isObject(options)) {
             return;
         }
-        _.has(options, 'server') && this.server.update(options.server);
-        _.isString(options.match) && (this.match = new UrlMatchPattern(options.match));
-        _.has(options, 'server') && (this.tunnel = _.isBoolean(options.tunnel) ? options.tunnel : false);
+        _.isString(options.host) && (this.host = options.host.replace(regexes.protocolMatcher, ''));
```
czardoz: If you're doing this, why are you assigning a default value separately above? Check some other constructors like `Url` etc to see how to do this right.

lib/url-pattern/url-match-pattern.js
```diff
@@ -51,7 +54,17 @@ _.assign(UrlMatchPattern.prototype, /** @lends UrlMatchPattern.prototype */ {
          * The url match pattern string
          * @type {String}
          */
-        this.pattern = _.get(options, 'pattern') || MATCH_ALL_URLS;
+        this.pattern = _.get(options, 'pattern') || this.pattern || MATCH_ALL_URLS;
```
czardoz: This means that we can never unset a pattern using the `.update()` method. :scream:

lib/url-pattern/url-match-pattern.js
```diff
@@ -51,7 +54,17 @@ _.assign(UrlMatchPattern.prototype, /** @lends UrlMatchPattern.prototype */ {
...
+
+        /**
+         * The protocols which will be used to match the URLs
+         * @type {Array}
+         */
+        this.protocols = _.get(options, 'protocols', this.protocols) || DEFAULT_PROTOCOLS;
+        !_.isArray(this.protocols) && (this.protocols = DEFAULT_PROTOCOLS);
+
+        this._allowProtocols = _.get(options, 'allowProtocols', this._allowProtocols) || false;
```
czardoz: Can you add a comment explaining what this is and why it's there?

lib/url-pattern/url-match-pattern.js
```diff
@@ -64,19 +77,26 @@ _.assign(UrlMatchPattern.prototype, /** @lends UrlMatchPattern.prototype */ {
      */
     createMatchPattern: function () {
         var matchPattern = this.pattern,
+            match;
+
+        // Check the match pattern of sanity and split it into protocol, host and path
 
-            // Check the match pattern of sanity and split it into protocol, host and path
+        if (this._allowProtocols) {
```
czardoz: We should follow a consistent style. I see above that you've used a lot of inlining of conditionals. Why not do the same here?

lib/url-pattern/url-match-pattern.js
```diff
@@ -64,19 +77,26 @@ _.assign(UrlMatchPattern.prototype, /** @lends UrlMatchPattern.prototype */ {
...
         if (!match) {
             // This ensures it is a invalid match pattern
             return undefined;
         }
 
         return {
-            protocol: match[1],
-            host: match[2],
-            path: this.globPatternToRegexp(match[3])
+            protocol: this._allowProtocols ? null : match[1],
```
czardoz: Create whatever `var`s you need above and use them once. The `_allowProtocols` flag should only be used _once_ in the entire function, not in 4 separate branches (1 in the `if-else` and 3 ternary operators).

e.g: 
```
var protocol

if (_allowblabla) {
    protocol = something
}
else {
    protocol = somethingelse;
}

return { protocol: protocol, ....}
```

lib/url-pattern/url-match-pattern.js
```diff
@@ -103,11 +123,13 @@ _.assign(UrlMatchPattern.prototype, /** @lends UrlMatchPattern.prototype */ {
     testProtocol: function (protocol) {
         var matchRegexObject = this._matchPatternObject;
 
-        // Do not match if protocol is invalid
+        if (this._allowProtocols) {
+            return _.indexOf(this.protocols, protocol) !== -1;
```
czardoz: Use a better function, like `_.contains`

czardoz:
@harryi3t Also add an entry to the changelog, like @shamasis did in #394

lib/collection/proxy-config.js
harryi3t: @SamvelRaja ^

lib/url-pattern/url-match-pattern.js
harryi3t: Coz here I needed an `else` block also. All the above cases are just `if-then`

Using that pattern here will make it very un-readable
```js
(this._allowProtocols && (match = matchPattern.match(regexes.noProtocolPatternSplit);) )
 || (match = matchPattern.match(regexes.patternSplit);)
```

lib/url-pattern/url-match-pattern.js
harryi3t: will use _.includes since _.contains is not there in `lodash` but in underscore.

lib/collection/proxy-config.js
harryi3t: done

lib/collection/proxy-config.js
SamvelRaja: @czardoz  Much more cleaner.

lib/collection/proxy-config.js
```diff
@@ -52,18 +63,30 @@ _.inherit((
         ProxyConfig.super_.call(this, options);
 
         // Assign defaults before proceeding
-        _.assign(this, /** @lends ProxyConfig.prototype */ {
+        _.assign(this, /** @lends ProxyConfig */ {
             /**
              * The url mach for which the proxy has been associated with.
              * @type {String}
              */
             match: new UrlMatchPattern(),
```
SamvelRaja: Pass the default protocol 😢

lib/collection/proxy-config.js
```diff
@@ -85,9 +116,23 @@ _.assign(ProxyConfig.prototype, /** @lends ProxyConfig.prototype */ {
...
+        _.isBoolean(options.tunnel) && (this.tunnel = options.tunnel);
+
+        // protocols
+        _.isArray(options.protocols) && (this.protocols = options.protocols) ||
+        _.isString(options.protocols) && (this.protocols = [options.protocols]);
+
+        // match pattern
+        if (_.isString(options.match)) {
+            var matchPatternOptions = {
+                allowProtocols: true,
```
SamvelRaja: Remove pls.

lib/collection/proxy-config-list.js
```diff
@@ -16,8 +16,8 @@ _.inherit((
      * @example <caption>Create a new ProxyConfigList</caption>
      * var ProxyConfigList = require('postman-collection').ProxyConfigList,
      *     myProxyConfig = new ProxyConfigList({}, [
-     *              {match: 'https://example.com/*',server: 'https://proxy.com',tunnel: true},
-     *              {match: 'https://example2.com/*',server: 'http://proxy2.com'},
+     *              {match: 'example.com/*',host: 'proxy.com', port: 8080, protocols: ['https'], tunnel: true},
```
shamasis: whitespace missing at `,host`

lib/collection/proxy-config.js
```diff
@@ -52,18 +63,30 @@ _.inherit((
...
+             * @type {Integer}
              */
-            server: new Url(),
+            port: DEFAULT_PORT,
+
+            /**
+             * The protocol(s) that the proxy server supports
+             * @type {Array}
+             */
+            protocols: DEFAULT_PROTOCOLS,
```
shamasis: This will cause the `DEFAULT_PROTOCOLS` to get mutated. Do a clone of the same or `DEFAULT_PROTOCOLS.splice()`

lib/collection/proxy-config.js
```diff
@@ -85,9 +116,22 @@ _.assign(ProxyConfig.prototype, /** @lends ProxyConfig.prototype */ {
         if (!_.isObject(options)) {
             return;
         }
-        _.has(options, 'server') && this.server.update(options.server);
-        _.isString(options.match) && (this.match = new UrlMatchPattern(options.match));
-        _.has(options, 'server') && (this.tunnel = _.isBoolean(options.tunnel) ? options.tunnel : false);
+        _.isString(options.host) && (this.host = options.host.replace(regexes.protocolMatcher, ''));
+        _.isInteger(options.port) && (options.port >= MIN_PORT && options.port <= MAX_PORT) && (this.port = options.port);
```
shamasis: Do not add port min max logic here. As a rest client, we should allow all port values.

lib/url-pattern/url-match-pattern.js
```diff
@@ -64,19 +80,27 @@ _.assign(UrlMatchPattern.prototype, /** @lends UrlMatchPattern.prototype */ {
      */
     createMatchPattern: function () {
         var matchPattern = this.pattern,
+            allowProtocols = _.isArray(this.protocols),
```
shamasis: Instead of all these, why not give a getter function `.protocol` which returns the protocol from the match itself?

PR #398, opened 2017-06-27
Feature/fix proplist remove

czardoz:
Fixes the bad behavior of blanket removal of value array when any key matches.

### DIFF
--- a/lib/collection/property-list.js
+++ b/lib/collection/property-list.js
@@ -215,7 +215,24 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
             if (predicate.apply(context, arguments)) {
                 if ((index = item[this._postman_listIndexKey]) && (index = String(index))) {
                     this._postman_listIndexCaseInsensitive && (index = index.toLowerCase());
-                    delete this.reference[index];
+
+                    if (this._postman_listAllowsMultipleValues && _.isArray(this.reference[index])) {
+                        // since we have an array of multiple values, remove only the value for which the
+                        // predicate returned truthy. If the array becomes empty, just delete it.
+                        _.remove(this.reference[index], function (each) {
+                            return each === item;
+                        });
+
+                        // If the array becomes empty, remove it
+                        (this.reference[index].length === 0) && (delete this.reference[index]);
+
+                        // If the array contains only one element, remove the array, and assign the element
+                        // as the reference value
+                        (this.reference[index].length === 1) && (this.reference[index] = this.reference[index][0])
+                    }
+                    else {
+                        delete this.reference[index];
+                    }
                 }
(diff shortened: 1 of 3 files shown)


lib/collection/property-list.js
```diff
@@ -215,7 +220,24 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
+
+                    if (this._postman_listAllowsMultipleValues && _.isArray(this.reference[index])) {
+                        // since we have an array of multiple values, remove only the value for which the
+                        // predicate returned truthy. If the array becomes empty, just delete it.
+                        _.remove(this.reference[index], function (each) {
+                            return each === item;
+                        });
+
+                        // If the array becomes empty, remove it
+                        (this.reference[index].length === 0) && (delete this.reference[index]);
```
kunagpal: Can be replaced with `_.isEmpty(this.reference[index]) && (delete this.reference[index]);`

test/unit/property-list.test.js
```diff
@@ -630,6 +631,97 @@ describe('PropertyList', function () {
...
+            // Ensure that the reference array contains an array of values for `key1`
+            expect(list.reference.key1).to.be.an(Array);
+            expect(list.reference.key1).to.have.length(2);
+
+            list.remove(function (fake) {
+                return fake.keyAttr === 'key1';
+            });
+
+            // should have removed the array of values
+            expect(list.reference.key1).to.be(undefined);
```
kunagpal:
```js
expect(list.reference).to.not.have.property('key1');
```

lib/collection/property-list.js
```diff
@@ -196,14 +197,18 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
-            (match = this.one(predicate)) && (predicate = function (item) {
-                return (item === match);
-            });
+            // if predicate is id, then create a function to remove that
+            // need to take care of case sensitivity as well :/
+            match = this._postman_listIndexCaseInsensitive ? predicate.toLowerCase() : predicate;
+            predicate = function (item) {
+                var id = item[this._postman_listIndexKey];
+                this._postman_listIndexCaseInsensitive && (id = id.toLowerCase());
+                return (id === match);
```
kunagpal: Unnecessary parentheses

lib/collection/property-list.js
czardoz: No need for unnecessary lodash usage, it does additional things. We're guaranteed to have an array because of the condition above, so directly checking the length is safe and extremely fast.

kunagpal:
Note to self: The predicate here comes in as either a string or an instance to search by for removal. This is being changed to a function every time, which will incur performance hits due to V8 optimizations being dropped for `.remove()`. Determine the magnitude of this hit, profile it, etc.

PR #403, opened 2017-07-04
Fix a typo in variable name

harryi3t:


lib/collection/proxy-config.js
```diff
@@ -147,7 +147,7 @@ _.assign(ProxyConfig.prototype, /** @lends ProxyConfig.prototype */ {
         }
 
         updatedProtocols = _.intersection(AllOWED_PROTOCOLS, _.castArray(protocols));
-        _.isEmpty(updatedProtocols) && (updatedProtocols = ALLOWED_PROTOCOLS);
```
SamvelRaja: It is right, change in initializer

PR #410, opened 2017-07-11
Removed v2 collection schemas from source

kunagpal:


shamasis:
Instead of this, we should migrate all schema to this repo

PR #418, opened 2017-07-17
Update .travis.yml to kill ssh-agent

shamasis:


shamasis:
The change is already done in https://github.com/postmanlabs/postman-collection/pull/417

PR #428, opened 2017-08-01
Empty header bugfix

kunagpal:
Reported by @DotNetRockStar in https://github.com/postmanlabs/newman/issues/1154.


shamasis:
@kunagpal - it seems that in update, a blank key was sent. Which implies that it has a key and as such it should be updated. Then why are we changing it?

shamasis:
If it works in postman - why's it different for Newman? Please explain F2F.

shamasis:
The fix cannot go in the constructor or the update function since, collection while saved, can have a state where key is blank. Instead, the place where runtime (or anyone) procures the header for usage, the check is put in there. That would be in runtime > requester > util -> getRequestHeaders

Issue #415, opened 2017-07-13
Support "disabled" in query params


WoLfulus:
I'm trying to generate a collection with several `disabled by default` query parameters, but passing `disabled` to query params doesn't actually disables them.

Looking at the exported JSON and the source code, only `key` and `value`  are present.

Is there a reason why `disabled` wasn't included in query params like headers?

kunagpal:
This a fix for this bug has been added in #429, and has now been released. Feel free to reopen if you're facing difficulties. Thanks! 😄

PR #429, opened 2017-08-01
Corrected disabled Query Param handling

kunagpal:
With this pull request, a query param list with one or more disabled query params will be correctly converted to it's string equivalent.

This takes care of https://github.com/postmanlabs/newman/issues/1114, https://github.com/postmanlabs/postman-app-support/issues/3361 and #415

### DIFF
--- a/lib/collection/query-param.js
+++ b/lib/collection/query-param.js
@@ -150,18 +150,20 @@ _.assign(QueryParam, /** @lends QueryParam */ {
             }, []);
         }
 
-        var encode = options && options.encode,
-            ignoreDisabled = options && options.ignoreDisabled,
+        var str,
+            encode = options && options.encode,
+            ignoreDisabled = options && options.ignoreDisabled;
 
-            // construct a query parameter string from the list, with considerations for disabled values
-            str = params.reduce(function (result, param) {
-                // If disabled parameters are to be ignored, bail out here
-                if (ignoreDisabled && param.disabled) { return result; }
+        // construct a query parameter string from the list, with considerations for disabled values
+        str = params.reduce(function (result, param) {
+            // If disabled parameters are to be ignored, bail out here
+            if (ignoreDisabled && (param.disabled === true)) { return result; }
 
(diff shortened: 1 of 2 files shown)


kunagpal:
@shamasis I've compacted and cleaned the logic used here.

lib/collection/query-param.js
```diff
@@ -136,26 +136,34 @@ _.assign(QueryParam, /** @lends QueryParam */ {
...
-        str = params.map(function (param) {
-            return QueryParam.unparseSingle(param, encode);
-        }).join(AMPERSAND);
+        var encode = options && options.encode,
+            ignoreDisabled = options && options.ignoreDisabled,
+
+            // construct a query parameter string from the list, with considerations for disabled values
+            str = params.reduce(function (result, param) {
+                // If disabled parameters are to be ignored, bail out here
+                if (ignoreDisabled && param.disabled) { return result; }
```
shamasis: Disabled === true?

lib/collection/query-param.js
```diff
@@ -136,26 +136,34 @@ _.assign(QueryParam, /** @lends QueryParam */ {
...
         }
 
-        str = params.map(function (param) {
-            return QueryParam.unparseSingle(param, encode);
-        }).join(AMPERSAND);
+        var encode = options && options.encode,
+            ignoreDisabled = options && options.ignoreDisabled,
+
+            // construct a query parameter string from the list, with considerations for disabled values
+            str = params.reduce(function (result, param) {
```
shamasis: Keep str; in var and move the function body out of var definition. One level less indentation and that the main work of this function is then not tucked inside a var.

kunagpal: Done 👍

lib/collection/query-param.js
kunagpal: Done 👍

shamasis:
ALso, new algorithm is 4.7 times faster than old.

PR #436, opened 2017-08-11
Code coverage improvements - Part III

kunagpal:
* Before:
```
Statements   : 94.46% ( 1211/1282 )
Branches     : 91.16% ( 1166/1279 )
Functions    : 94.51% ( 310/328 )
Lines        : 94.56% ( 1182/1250 )
```

* After:
```
Statements   : 99.3% ( 1272/1281 )
Branches     : 97.81% ( 1251/1279 )
Functions    : 99.7% ( 327/328 )
Lines        : 99.36% ( 1241/1249 )
```

### DIFF
--- a/test/unit/response.test.js
+++ b/test/unit/response.test.js
@@ -21,6 +21,33 @@ describe('Response', function () {
             expect(response).to.be.ok();
         });
 
+        it('should handle the absence of Buffer gracefully', function () {
+            var response,
+                originalBuffer = Buffer;
+
+            delete global.Buffer;
+            expect(function () {
+                response = new Response({ stream: originalBuffer('random').toJSON() });
+            }).to.not.throwError();
+
+            expect(response).to.be.ok();
+            global.Buffer = originalBuffer;
+        });
+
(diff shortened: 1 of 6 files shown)


test/unit/response.test.js
```diff
@@ -21,6 +21,33 @@ describe('Response', function () {
             expect(response).to.be.ok();
         });
 
+        it('should handle the absence of Buffer gracefully', function () {
+            var response,
+                originalBuffer = Buffer;
+
+            delete global.Buffer;
+            expect(function () {
+                response = new Response({ stream: originalBuffer('random').toJSON() });
```
kamalaknn: Can you move the value assigned to `stream` up as a variable?

It would make more sense to read the test as 

1. Create a value as `Buffer('random').toJSON()`
2. Remove `Buffer`
3. Make sure `new Response()` works without `Buffer`
4. Restore `Buffer`

Right now, removing the `Buffer` and accessing it again looks weird.

kunagpal: Done.

PR #437, opened 2017-08-18
Better VariableScope layers

kunagpal:
With this pull request, the `_layers` property is only initialized to a blank array on demand. It was an empty array by default earlier.

Code coverage remains at 98%, same as before.

### DIFF
--- a/lib/collection/variable-scope.js
+++ b/lib/collection/variable-scope.js
@@ -82,9 +82,9 @@ _.inherit((
         // in above line, we clone the values if it is already a list. there is no point directly using the instance of
         // a variable list since one cannot be created with a parent reference to begin with.
 
-        this._layers = [];
-
         if (layers) {
+            this._layers = [];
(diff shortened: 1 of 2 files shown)


lib/collection/variable-scope.js
```diff
@@ -82,9 +82,9 @@ _.inherit((
         // in above line, we clone the values if it is already a list. there is no point directly using the instance of
         // a variable list since one cannot be created with a parent reference to begin with.
 
-        this._layers = [];
-
         if (layers) {
+            this._layers = [];
```
shamasis: I recall already doing this. How did it get reverted?

kunagpal: Looks like that was my fault 😓 
https://github.com/postmanlabs/postman-collection/pull/384/commits/851ca618f399a93926f56ba3fa7209e6e5473022

PR #438, opened 2017-08-18
Fixed AWS Auth bug where signing failed if URL had port

harryi3t:


lib/collection/request-auth/awsv4.js
```diff
@@ -55,7 +55,7 @@ module.exports = {
                 secretAccessKey: params.secretKey,
                 sessionToken: params.sessionToken || undefined
             },
-            host: request.url.getHost(),
+            host: request.url.getRemote(),
```
harryi3t: adds the port number to the host if present (does not add the default implicit ports like 80 and 443 if not present in the URL)
e.g.
```
www.google.com -> google
www.foo.google.com -> foo.google
www.foo.google.com:80 -> foo.google:80
```

PR #442, opened 2017-08-23
Polymorphic raw request bodies

kunagpal:
- [x] Polymorphic raw request bodies supported as strings and arbitrary objects
- [x] Objects of the form `{ mode: 'raw' }` remain unaffected.
- [x] JSDoc comments for `RequestBody~update` updated to reflect expanded scope.
- [x] Unit tests added

With this pull request, raw request bodies can be accepted in multiple forms, as shown below:
1. As a raw string
```js
var rBody = new sdk.RequestBody('This is meant to be a raw body');

rBody.mode // `raw`
rBody.raw // `This is meant to be a raw body`
```

2. As an arbitrary object
```js
var rBody = new sdk.RequestBody({ foo: 'bar' });

rBody.mode // `raw`
rBody.raw // { foo: 'bar' }
```


lib/collection/request-body.js
```diff
@@ -17,26 +17,26 @@ _.inherit((
...
      *
-     * @param {Object} options
+     * @param {*} options
      */
     update: function (options) {
-        if (!options.mode) { return; } // need a valid mode @todo raise error?
+        if (!options) { return; }
 
-        var mode = RequestBody.MODES[options.mode.toString().toLowerCase()] || RequestBody.MODES.raw,
+        var rawMode = options.mode && options.mode.toString().toLowerCase(),
```
shamasis: `rawMode` is ambiguous name as it seems one is talking about "raw" mode. This variable is not needed since we can do:

```
mode = RequestBody.MODES[options.mode && options.mode.toString().toLowerCase()] || RequestBody.MODES.raw,
```

test/unit/request-auth.test.js
```diff
@@ -438,7 +438,7 @@ describe('RequestAuth', function () {
                 },
                 certificate: undefined,
                 proxy: undefined,
-                body: { },
+                body: { mode: 'raw', raw: { foo: 'bar' } },
```
shamasis: change this test to only look into auth. Why look at entire request?

lib/collection/request-body.js
```diff
@@ -17,26 +17,26 @@ _.inherit((
      * @constructor
      * @extends {PropertyBase}
      *
-     * @param {Object} options
+     * @param {*} options
      */
     RequestBody = function PostmanRequestBody (options) {
```
shamasis: This should work like headers in some way. If `options` is specifically string then we modify options to:

```
options = {
    mode: 'raw',
    raw: options
}

File: lib/collection/request-body.js [hunk 4]
```diff
@@ -70,6 +80,7 @@ _.assign(RequestBody.prototype, /** @lends RequestBody.prototype */ {
         // If mode is urlencoded but options does not provide any content, set it to an empty property list
         (mode === RequestBody.MODES.formdata && !formdata) && (formdata = new PropertyList(FormParam, this, []));
 
+        // @todo: Possibly use _.mergeDefined here?
```
shamasis: not yet. since it allows us to remove other modes' data for the time being. will not know the impact of this change. also, if we assume that this behaviour is expected, then the todo can be removed.

lib/collection/request-body.js
```diff
@@ -61,6 +61,16 @@ _.assign(RequestBody.prototype, /** @lends RequestBody.prototype */ {
 
         _.isString(options.file) && (file = { src: file });
 
+        if (!raw) {
```
shamasis: why is this needed? + line 65 does close to this, followed subsequent removal of raw in the _.assign below

lib/collection/request-body.js
kunagpal: Yes, that is the current behaviour.

test/unit/request-auth.test.js
kunagpal: If I recall correctly, this test was failing due to the raw body behaviour change. I'll revisit.

lib/collection/request-body.js
kunagpal: This handles raw bodies regardless of the current request body mode. The condition below is to sanitize raw bodies only if a raw body is currently in use.

PR #446, opened 2017-08-27
Removal of auth related functionalities

shamasis:
This does the following:

1. Completely removes all individual auth code
2. Equips PropertyList to deal with add-or-insert operations
3. Converts RequestAuth to deal with arbitrary `type` of auth and store its data in VariableList
4. RequestAuth constructor still accepts object type auth parameters, but toJSON spits out Array

> Pending a way to address schema representation and README changes

PS: Since I ran into some dependency mismatch in package.json, I added a system test to ensure the requires in code are reflected in package.json


czardoz:
Looks good, should be a major version bump

Also, need to update:

1. `postman-collection-transformer`
2. `postman-schemas`
3. The App has code which maps older (v1) parameters to new ones and vice versa, so that needs to be updated to use arrays now
4. The "update request" functionality in the app needs to be updated as well (or removed)

:+1:

lib/collection/property-list.js
```diff
@@ -190,6 +190,33 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
     /**
+     * Add an item or update an existing item
+     *
+     * @param {PropertyList~Type} item
+     *
+     * @returns {?Boolean}
+     */
+    upsert: function (item) {
+        // do not proceed on empty param, but empty strings are in fact valid.
+        if (_.isNull(item) || _.isUndefined(item) || _.isNaN(item)) { return null; }
```
kunagpal:
```js
if (_.isNil(item) || _.isNaN(item)) { return null; }
```

lib/collection/request-auth.js
```diff
@@ -131,122 +23,96 @@ _.inherit((
...
      */
     update: function (options, type) {
-        type = (type ? type.toString().toLowerCase() : this.type);
-        // procure the handler of specific type and then call it's update method with the parameter definition as it's
-        // scope.
-        // @todo raise error on invalid type?
-        var AuthType = RequestAuth.types[type];
-        AuthType && (this[type] = new AuthType(options));
+        // update must have options
+        if (!options || !_.isObject(options)) { return; }
```
kunagpal: `_.isObject` can be used directly.

lib/collection/property-list.js
```diff
@@ -289,6 +316,40 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
 
     /**
+     * Add or update values from a source list.
+     *
+     * @param {PropertyList|Array} source
+     * @param {Boolean=false} prune Setting this to `true` will cause the extra items from the list to be deleted
+     */
+    assimilate: function (source, prune) {
+        if (PropertyList.isPropertyList(source)) {
+            source = source.members;
```
kunagpal: It would be nice to assign source.members to a fresh variable instead of changing the type on the fly.

lib/collection/request-auth.js
```diff
@@ -131,122 +23,96 @@ _.inherit((
...
 
 _.assign(RequestAuth.prototype, /** @lends RequestAuth.prototype */ {
     /**
      * Update the parameters of a specific authentication type. If none is provided then it uses the one marked as to be
      * used.
      *
-     * @param {Object} options
-     * @param {AuthTypes=} [type]
+     * @param {VariableList|Array|Object} options
+     * @param {String=this.type} [type]
```
harryi3t: I might be wrong here but, isn't the correct way is 
```js
// @param {String} [type=this.type]
```
http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

lib/collection/request-auth.js
```diff
@@ -131,122 +23,96 @@ _.inherit((
...
-        var AuthType = RequestAuth.types[type];
-        AuthType && (this[type] = new AuthType(options));
+        // update must have options
+        if (!options || !_.isObject(options)) { return; }
+        // validate type parameter and/or choose default from existing type.
+        if (!_.isString(type) || (type === 'type')) { type = this.type; }
+
+        var parameters = this[type];
+
+        // in case the type holder is not created, we create one and send the population varuables
```
harryi3t: typo `variables`

lib/collection/property-list.js
shamasis: Can you elaborate @kunagpal ?

kunagpal: Within the `if` block, `source` is an instance of `PropertyList`, which then gets changed to an `Array`. This will cause an optimization bailout within the `assimilate` function.

lib/collection/request-auth.js
```diff
@@ -1,122 +1,14 @@
...
-    },
+    VariableList = require('./variable-list').VariableList,
 
     RequestAuth;
 
 /**
+ * This defines the definition of the authentication method to be used.
+ *
  * @typedef RequestAuth~definition
  * @property {String=} type The Auth type to use. Check the names in {@link AuthTypes}
```
kamalaknn: @shamasis How do we document

```
@property {VariableList} awsv4
@property {VariableList} basic
@property {VariableList} digest
```

lib/collection/request.js
```diff
@@ -107,6 +107,38 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
+     * @param {?String|RequestAuth~definition} type
+     * @param {VariableList~definition=} [options]
+     *
+     * @note This function was previously (in v2 of SDK) used to clone request and populate headers. Now it is used to
+     * only set auth information to request
+     *
+     * @note that ItemGroup#authorizeUsing depends on this function
+     */
+    authorizeUsing: function (type, options) {
+        if (_.isObject(type) && _.isNil(options)) {
```
kamalaknn: @shamasis should this be `_.isObject(type) && !_.isNil(options)`

lib/collection/request.js
```diff
@@ -107,6 +107,38 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
+     *
+     * @param {?String|RequestAuth~definition} type
+     * @param {VariableList~definition=} [options]
+     *
+     * @note This function was previously (in v2 of SDK) used to clone request and populate headers. Now it is used to
+     * only set auth information to request
+     *
+     * @note that ItemGroup#authorizeUsing depends on this function
+     */
+    authorizeUsing: function (type, options) {
```
kamalaknn: @shamasis Two points,

1. This API looks weird, we would have to do 
```
item.request.authorizeUsing(item.getAuth());
auth = item.getAuth();
```
`auth` is set on a request, but get from an item. `authorizeUsing` does not return the `auth`. So having to do `context.item.getAuth()` again.

2. We set the parent of the auth to the request, but the parent of request is not set to the item 😕

Can we do `auth.setParent(this)` in `item.getAuth` ?

lib/collection/request.js
shamasis: Done

lib/collection/request.js
shamasis: 1. This function is meant to set auth in a group or item. It is not for getting auth from somewhere to set here. It is more like functions such as .setHeaders, etc.

2. Then we should bypass request and set parent to item. That would be weird though.

kamalaknn: Can we set the parent of the request to the item?

shamasis: Added reference resolution to `.toObjectResolved` instead.

lib/collection/item.js
```diff
@@ -136,17 +136,12 @@ _.assign(Item.prototype, /** @lends Item.prototype */ {
     /**
      * Fetches applicable AuthType from the current item.
      *
-     * returns {RequestAuthBase|undefined}
+     * returns {RequestAuth}
```
harryi3t:
```js
returns {RequestAuth|undefined}
```

lib/collection/request.js
```diff
@@ -245,12 +277,12 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
      * Creates a copy of this request, with the appropriate auth headers or parameters added.
      *
+     * @deprecated discontinued in v3.x
      * @note This function does not take care of resolving variables.
      * @returns {Request}
      */
     authorize: function () {
-        var clonedReq = this.clone();
-        return RequestAuth.authorize(clonedReq);
+        throw new Error('collection request.authrozie() has been discontinued');
```
kunagpal: typo

lib/collection/property-list.js
```diff
@@ -289,6 +316,37 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
+        members.forEach(function (item) {
+            if (!(item && item.hasOwnProperty(indexer))) { return; }
+            list.upsert(item);
+            sourceKeys[item[indexer]] = true;
+        });
+
+        // now remove any variable that is not in source object
+        // @note - using direct `this.reference` list of keys here so that we can mutate the list while iterating
+        // on it
+        prune && _.forEach(_.keys(list.reference), function (key) {
```
kunagpal: `_.forEach` works well with objects as well, the extra `_.keys` call can be avoided.

lib/collection/variable-list.js
```diff
@@ -83,7 +99,7 @@ _.assign(VariableList.prototype, /** @lends VariableList.prototype */ {
         // now remove any variable that is not in source object
         // @note - using direct `this.reference` list of keys here so that we can mutate the list while iterating
         // on it
-        _.forEach(_.keys(list.reference), function (key) {
+        (prune !== false) && _.forEach(_.keys(list.reference), function (key) {
```
kunagpal: Similar to the comment above.

PR #472, opened 2017-09-23
Add exception to nsprc for https://nodesecurity.io/advisories/531

prakharjoshi:


shamasis:
It is a very common advisory. Please do not add trrse to other repos. Instead follow https://github.com/postmanlabs/postman-collection/pull/469 and add specific package name to it.

PR #494, opened 2017-11-02
Add `ignoreOwnVariables` option to `Property.prototype.toObjectResolved`

kamalaknn:
Allows `toObjectResolved` to ignore own variables when used with an external source of variable definitions.

### DIFF
--- a/lib/collection/property.js
+++ b/lib/collection/property.js
@@ -141,36 +141,54 @@ _.assign(Property.prototype, /** @lends Property.prototype */ {
     /**
      * Returns an object representation of the Property with its variable references substituted.
      *
+     * @example <caption>Resolve an object using variable definitions from itself and its parents</caption>
+     * property.toObjectResolved();
+     *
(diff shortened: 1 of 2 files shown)


lib/collection/property.js
```diff
@@ -141,37 +141,55 @@ _.assign(Property.prototype, /** @lends Property.prototype */ {
...
      * @throws {Error} If `variables` cannot be resolved up the parent chain.
      */
-    toObjectResolved: function (scope, overrides) {
-        // 1. if variables is passed as params, use it or fall back to oneself
-        // 2. for a source from point (1), and look for `.variables`
-        // 3. if `.variables` is not found, then rise up the parent to find first .variables
-
-        var variableSourceObj = scope || this,
+    toObjectResolved: function (scope, overrides, options) {
+        var useOwnVariablesToResolve = options && options.useOwnVariablesToResolve,
```
kunagpal: The option name is too verbose :/

PR #502, opened 2017-11-08
Add support for system property in query parameters

harryi3t:
This will be used by runtime in auth for setting query parameters

### DIFF
--- a/lib/collection/query-param.js
+++ b/lib/collection/query-param.js
@@ -59,6 +59,7 @@ _.assign(QueryParam.prototype, /** @lends QueryParam.prototype */ {
             key: _.get(param, 'key'), // we do not replace falsey with blank string since null has a meaning
             value: _.get(param, 'value')
         });
+        _.has(param, 'system') && (this.system = param.system);
     },
 
     valueOf: function () {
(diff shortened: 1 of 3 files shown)


lib/collection/query-param.js
```diff
@@ -59,6 +59,7 @@ _.assign(QueryParam.prototype, /** @lends QueryParam.prototype */ {
             key: _.get(param, 'key'), // we do not replace falsey with blank string since null has a meaning
             value: _.get(param, 'value')
         });
+        _.has(param, 'system') && (this.system = param.system);
```
kunagpal: This will also include falsy `system` values, which is not what we want. Alternatively,
```js
param && param.system && (this.system = true);
```

kamalaknn: @kunagpal if we do that, there would be no way of unsetting a `system:true` value 🤔

kunagpal: In that case, there should be a check to remove `system` if it is falsy, like in the original snippet.

harryi3t: @kunagpal But that's what we are doing at other places as well
```js
// variable.js
options.hasOwnProperty('system') && (this.system = options.system);

// header.js
_.has(options, 'system') && (this.system = options.system);
```

All of these will go away once `Property` itself starts supporting this property.
We should keep consistency in the behaviour at all the places where this is used.

PR #512, opened 2017-11-15
Make sure `item#getAuth` continues parent lookup even if a match with empty auth is found

kamalaknn:


lib/collection/property-base.js
```diff
@@ -58,38 +58,39 @@ _.assign(PropertyBase.prototype, /** @lends PropertyBase.prototype */ {
...
-        if (this[property]) {
-            return this;
-        }
-
-        // Start travelling up the parent chain, to find the required property.
-        parent = this.__parent;
-
-        while (parent) {
-            if (parent[property]) {
+        do {
```
shamasis: two loops @harryi3t

lib/collection/property-base.js
```diff
@@ -58,38 +58,39 @@ _.assign(PropertyBase.prototype, /** @lends PropertyBase.prototype */ {
...
+            // if customizer is present test with it
+            // else check for existence
+            if (customizer ? customizer.call(this, parent) : parent[property]) {
```
shamasis: outside loop do `customizer = customiser.bind(this)`

PR #513, opened 2017-11-15
Ignore falsy `auth` values in `new {Request,ItemGroup,Collection}`

kamalaknn:


kamalaknn:
cc @SamvelRaja

kamalaknn:
@shamasis didn't have to update `authorizeUsing`,

* `null` is designed to remove auth
* falsy `type` is ignored

lib/collection/request.js
```diff
@@ -93,7 +93,9 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
             /**
              * @type {RequestAuth}
              */
-            auth: _.createDefined(options, 'auth', RequestAuth),
+            // auth is a special case, empty RequestAuth should not be created for falsy values
```
shamasis: inline comment above or inside type as note. @harryi3t

PR #515, opened 2017-11-18
Introduced `json` type for variables

kamalaknn:
**Changes**
* Added a `bi-directional` type casting functions to support custom types (@shamasis)
* Fixed a bug where `type` updates were removing `value` functions
* Improved tests around variable types in initializing and casting

**Trivia**
* `variable.set(JSON.stringify(object)) === variable.set(object)`
* `variable.valueType('json'); variable.toString() === '[object Object]'`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,21 @@
 # Postman Collection SDK Changelog
 
 #### Unreleased
+* :tada: Get and set objects in variables with `json` variable type
+```js
+var variable = new Variable({ type: 'json' });
+
+// set objects as values for `json types`
+variable.set({ version: 'v1' });
+
+// get object values
+console.log(variable.get()); // { version: 'v1' }
+console.log(typeof variable.get()); // object
+```
 * #502 Added support for system property in query parameters
 * #503 `request~toJSON` no longer converts `url` to string
 * Made sure all Script instances have an id
+* :bug: Fixed a bug where updating types in a variable removed value functions
 * :bug: `auth` is now included in `collection.toJSON`
 * :bug: Fixed a bug where `item.getAuth` would return an empty `RequestAuth`, instead of looking up in the parent
 * :bug: Fixed a bug where empty `auth` was created for `new {Request,ItemGroup,Collection}({auth: null})`. This affected the auth lookup in parents.
--- a/lib/collection/variable.js
+++ b/lib/collection/variable.js
@@ -75,7 +75,7 @@ _.assign(Variable.prototype, /** @lends Variable.prototype */ {
      * @returns {Variable.types}
      */
     get: function () {
-        return _.isFunction(this.value) ? this.castOut(this.value()) : this.castOut(this.value);
+        return this.castOut(this.value);
     },
 
     /**
@@ -268,7 +268,7 @@ _.assign(Variable, /** @lends Variable */ {
                     val = null;
                 }
 
-                return JSON.parse(val);
+                return val;
             }
         },
 
--- a/lib/collection/variable.js
+++ b/lib/collection/variable.js
@@ -153,10 +153,12 @@ _.assign(Variable.prototype, /** @lends Variable.prototype */ {
      */
     valueType: function (typeName, _noCast) {
         !_.isNil(typeName) && (typeName = typeName.toString().toLowerCase()); // sanitize
-
         if (!Variable.types[typeName]) {
             return this.type || ANY; // @todo: throw new Error('Invalid variable type.');
         }
+
+        // set type if it is valid
+        this.type = typeName;
 
         // 1. get the current value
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,21 @@
 # Postman Collection SDK Changelog
 
 #### Unreleased
+* :tada: Get and set objects in variables with `json` variable type
+```js
+var variable = new Variable({ type: 'json' });
+
+// set objects as values for `json types`
+variable.set({ version: 'v1' });
```
shamasis: the 2nd arg needs to be `json` --- isn't it?

lib/collection/variable.js
```diff
@@ -245,7 +246,8 @@ _.assign(Variable, /** @lends Variable */ {
              */
             in: function (val) {
                 try {
-                    val = JSON.stringify(val);
+                    // @todo: shoud we check if `val` is a valid JSON string?
+                    val = typeof val === STRING ? val : JSON.stringify(val);
```
shamasis: why's this allowed?

lib/collection/variable.js
```diff
@@ -198,15 +237,64 @@ _.assign(Variable, /** @lends Variable */ {
...
+         */
+        'json': {
+            /**
+             * @param {Object|Array} val
+             * @returns {String}
+             */
+            in: function (val) {
+                try {
+                    // @todo: shoud we check if `val` is a valid JSON string?
+                    val = typeof val === STRING ? val : JSON.stringify(val);
```
shamasis: is this done to accept serialised data? if it is to give polymorphic input to user... is that good?

lib/collection/variable.js
```diff
@@ -166,7 +166,8 @@ _.assign(Variable.prototype, /** @lends Variable.prototype */ {
         // 3. then set the interstitial value
         var interstitialCastValue;
 
-        if (!_noCast) {
+        // do not touch value functions
+        if (!(_noCast || _.isFunction(this.value))) {
```
shamasis: 👍

lib/collection/variable.js
```diff
@@ -115,26 +116,64 @@ _.assign(Variable.prototype, /** @lends Variable.prototype */ {
...
-        !_nocast && (this.value = this.cast(this.value));
 
-        return typeName;
+        // 1. get the current value
+        // 2. set the new type if it is valid and cast the stored value
+        // 3. then set the interstitial value
+        var interstitialCastValue;
+
+        // do not touch value functions
+        if (!(_noCast || _.isFunction(this.value))) {
```
shamasis: 👍

lib/collection/variable.js
```diff
@@ -115,26 +116,64 @@ _.assign(Variable.prototype, /** @lends Variable.prototype */ {
...
-        if (!typeName || !Variable.types[(typeName = typeName && typeName.toString().toLowerCase())]) {
+    valueType: function (typeName, _noCast) {
+        !_.isNil(typeName) && (typeName = typeName.toString().toLowerCase()); // sanitize
+        if (!Variable.types[typeName]) {
             return this.type || ANY; // @todo: throw new Error('Invalid variable type.');
         }
 
-        // set the new type if it is valid and cast the stored value
+        // set type if it is valid
         this.type = typeName;
```
shamasis: damn... missed it. great find @kamalaknn

CHANGELOG.md
kamalaknn: `variable.set` only takes the value as a parameter. `type` is managed separately(at construction of using `valueType`)

lib/collection/variable.js
kamalaknn: This is done because otherwise JSON strings will be stringified again and stored. We would try to parse it and it would fail always.

So not doing this is more work while setting and getting, with no benefit.

lib/collection/variable.js
```diff
@@ -115,26 +116,64 @@ _.assign(Variable.prototype, /** @lends Variable.prototype */ {
...
+
+    /**
+     * Typecasts a value to the {@link Variable.types} of this {@link Variable}. Returns the value of the variable
+     * converted to the type specified in {@link Variable#type}.
+     *
+     * @private
+     * @param {*} value
+     * @returns {*}
+     */
+    castIn: function (value) {
```
kunagpal: Extremely similar to `castOut`, could be wrapped around a generic `cast` function.

shamasis: I’m afraid that wouldn’t be possible. There’s already a cast function which is public and that is exactly same as .castOut

Also - this is very low level function. It shouldn’t hurt and may even give more performance.

CHANGELOG.md
shamasis: 👍🏼

PR #523, opened 2017-11-27
Added `customizer` to `PropertyList.prototype.one`

kamalaknn:


kamalaknn:
Will revisit this when solving the variable resolution cleanly. Will need this for dropping disabled values.

PR #539, opened 2018-01-04
toString with undefined causes exceptions

stephan-nordnes-eriksen:
If the `value` is undefined, it would throw error on the `_.isFunction` test.

Not sure if this is a common state to get in, but it just happened to me.


stephan-nordnes-eriksen:
I might have made a mistake. This will cause all falsy values to fail. I believe it should actually test for equality to undefined. Eg. 'value === undefined'

I'm on my mobile phone only right now, so it is a bit hard to make that change.

stephan-nordnes-eriksen:
@shamasis ^

PR #546, opened 2018-01-07
Faster QueryParam unparse

kunagpal:
In cases where the input is a hash of query params, the stringified result is now directly computed and returned instead of deriving it from an intermediate array of objects state. This prevents double traversal on objects and avoids the associated performance hit.

Associated unit test: https://github.com/postmanlabs/postman-collection/blob/feature/faster-qp-hash-unparse/test/unit/query-param.test.js#L132


kamalaknn:
I'm seeing some differences in the behaviour, with encoding, `null` and `undefined` values, between object format and key value format, and between this branch and `develop`.

Before:
screen shot 2018-01-07 at 12 03 32 pm

After:
screen shot 2018-01-07 at 12 02 21 pm


Now, to make these consistent, we'll have to do it in both the places and that will give us not much of a perf boost.

shamasis:
@kamalaknn can you please add tests for that breaking change in a separate branch and send PR? I want to have that in develop and let this branch fail until that passes.

kamalaknn:
@shamasis done -> https://github.com/postmanlabs/postman-collection/pull/615

kamalaknn:
Did a quick perf check on the diff

screen shot 2018-04-26 at 1 05 22 pm

There's some repetition here, but I don't think we need to DRY it now. We have tests for both of these code flows. There's a 17% increase in performance after this change. That justifies having some repetition.

PR #589, opened 2018-03-02
Collection~toJSON description fix

kunagpal:
Discovered in https://community.getpostman.com/t/postman-sdk-not-adding-description-to-collection-info-block/1102

### DIFF
--- a/lib/collection/collection.js
+++ b/lib/collection/collection.js
@@ -195,7 +195,7 @@ _.assign(Collection.prototype, /** @lends Collection.prototype */ {
 
         // move ids and stuff from root level to `info` object
         json.info = {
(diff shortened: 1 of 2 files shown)


lib/collection/collection.js
```diff
@@ -195,7 +195,7 @@ _.assign(Collection.prototype, /** @lends Collection.prototype */ {
 
         // move ids and stuff from root level to `info` object
         json.info = {
-            id: this.id,
```
kamalaknn: Any reason why this was removed?

kunagpal: It's been replaced with `_posmtan_id`, the `id` field is not valid as per the `info` structure.

PR #607, opened 2018-03-19
Request body sanity fix

kunagpal:
cc @abhijitkane 

Closes #442

### DIFF
--- a/lib/collection/request.js
+++ b/lib/collection/request.js
@@ -83,9 +83,11 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
         _.has(options, 'method') && (this.method = _.isString(options.method) ? options.method.toUpperCase() :
             (options.method || DEFAULT_REQ_METHOD));
 
+        var body;
+
         if (_.has(options, 'body')) {
-            _.isString(options.body) && (options.body = { mode: 'raw', raw: options.body });
-            !_.isPlainObject(options.body) && (delete options.body);
+            body = options.body;
+            _.isString(body) && (body = { mode: 'raw', raw: body });
         }
 
         // The rest of the properties are not assumed to exist so we merge in the defined ones.
@@ -93,7 +95,7 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
             /**
              * @type {RequestBody|undefined}
              */
-            body: _.createDefined(options, 'body', RequestBody),
+            body: _.isPlainObject(body) ? new RequestBody(body) : undefined,
 
             // auth is a special case, empty RequestAuth should not be created for falsy values
             // to allow inheritance from parent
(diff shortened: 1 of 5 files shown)


lib/collection/request.js
```diff
@@ -83,6 +83,11 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
         _.has(options, 'method') && (this.method = _.isString(options.method) ? options.method.toUpperCase() :
             (options.method || DEFAULT_REQ_METHOD));
 
+        if (_.has(options, 'body')) {
+            _.isString(options.body) && (options.body = { mode: 'raw', raw: options.body });
```
kamalaknn: Not sure about the self-correction here.

lib/collection/request.js
```diff
@@ -83,6 +83,11 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
+            !_.isPlainObject(options.body) && (delete options.body);
```
kamalaknn: Why are we mutating options here?

lib/collection/request.js
kunagpal: As discussed offline.

lib/collection/request.js
```diff
@@ -83,17 +83,19 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
         _.has(options, 'method') && (this.method = _.isString(options.method) ? options.method.toUpperCase() :
             (options.method || DEFAULT_REQ_METHOD));
 
+        var body;
```
kamalaknn: Should be moved to top of the function.

lib/collection/request.js
```diff
@@ -83,17 +83,19 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
+
         if (_.has(options, 'body')) {
-            _.isString(options.body) && (options.body = { mode: 'raw', raw: options.body });
-            !_.isPlainObject(options.body) && (delete options.body);
+            body = options.body;
```
kamalaknn: `body = _.isString(body) ? { mode: 'raw', raw: body } : options.body;`

kunagpal: Before: 240K ops/s
After: 220K ops/s

test/unit/request-body.test.js
```diff
@@ -79,6 +79,18 @@ describe('RequestBody', function () {
...
     });
 
+    it('should work with string bodies correctly', function () {
+        var body = new RequestBody('foo');
+        expect(body).to.have.property('mode', 'raw');
+        expect(body).to.have.property('raw', 'foo');
+    });
+
+    it('should reject invalid body types', function () {
+        expect(new RequestBody([])).to.not.have.property('body');
```
kamalaknn: This will always pass 🤔

kunagpal: Yes. The collection SDK didn't validate for non-standard body options before.

kamalaknn: why are we asserting `new RequestBody().body`? this will always be empty right? even for valid body types?

kunagpal: Incorrect test; has been fixed.

PR #612, opened 2018-03-23
Update marked

joshbruce:
Given the scope of the PR - contributing guidelines weren't checked. Apologies if there is a missed convention.

[Marked](https://github.com/markedjs/marked) is in an active status again.

As part of due diligence I going around making sure project dependent on the `8fold-marked` workaround are updated. We have fixed a few defects including another discovered security vulnerability.


joshbruce:
One is glad to be of service. :)

PR #540, opened 2018-01-04
Added tests for PR#539

kamalaknn:

PR #625, opened 2018-04-18
Use toString for all variable values except null and undefined

kamalaknn:
We were converting all falsy values to empty strings as part of the PR #539.

The tests were added in PR #540 asserting that falsy values were stringified as empty strings. We had a verbal discussion about this behavior and were okay with that.

Turns out `postman-runtime` had tests that asserted falsy variable values were preserved during resolution.

I've updated the fix and added few more tests.


kunagpal:
@kamalaknn Can we have benchmarks on the before after?

kamalaknn:
Here's the benchmark for before and after the change.

screen shot 2018-04-19 at 2 41 25 pm

The existing implementation is fast because it is doing less conditional instructions but the actual difference between the two is just around 1% to 2% percent. The actual implementation for `_.isNil` is a simple `value == null`.

PR #629, opened 2018-04-25
Using filename from content disposition header on download of response

dhineshwiz:
Changes made to take file name from the content-disposition header and send the same in response

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "js-yaml": "3.11.0",
     "jsdoc": "3.5.5",
     "jsdoc-to-markdown": "4.0.1",
-    "karma": "2.0.0",
+    "karma": "2.0.2",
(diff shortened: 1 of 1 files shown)


test/unit/response.test.js
```diff
@@ -806,6 +806,22 @@ describe('Response', function () {
                 });
             });
 
+            it('Should take file name from content disposition header', function () {
```
kunagpal: A test should also be added for cases where matches is falsy.

kunagpal:
@dhineshwiz In addition to the comments above, we should also look into possible improvements in if-else branching.

kunagpal:
@dhineshwiz Use https://github.com/jshttp/content-disposition instead, it will save us the effort of accounting for every RFC compliant format.

PR #640, opened 2018-05-17
Fixing broken link to the Postman Collections page

ratpik:
Current URL redirects to https://www.getpostman.com/docs/v6/collections which does not exist


kamalaknn:
I would love to link to a docs page. I think the docs are versioned and the link can go out of date or broken soon. We should change to docs page again when we have a `/docs/latest` page.

PR #654, opened 2018-06-04
Update the in-repo schema to the latest 2.1 schema

kamalaknn:


shamasis:
@kunagpal ping

kunagpal:
@kamalaknn As discussed offline, we need to think of a way to retain documentation for auth parameters under the inbuilt helpers.

PR #657, opened 2018-06-07
Added support for tracking variable changes in VariableScopes

kamalaknn:
* Added `VariableChangeset`
* Added `VariableScopeDiff`
* Added `enableTracking` flag to `VariableScope`
* Added `VariableScope#enableTracking` and `VariableScope#disableTracking`

Will add a proper changelog if there are no API changes requested.

### DIFF
--- a/lib/collection/variable-changeset.js
+++ b/lib/collection/variable-changeset.js
@@ -1,0 +1,235 @@
+var _ = require('../util').lodash,
+    Property = require('./property').Property,
+
+    /**
+     * @private
+     * @const
+     * @type {Object}
+     */
+    INSTRUCTION = {
+        SET: 'set',
+        UNSET: 'unset'
+    },
+
+    /**
+     * @private
+     * @const
+     * @type {String}
+     */
+    POSTMAN_PROPERTY_NAME = '_postman_propertyName',
+
+    /**
+     * @private
+     * @const
+     * @type {String}
+     */
+    VARIABLE_SCOPE = 'VariableScope',
+
+    /**
+     * Checks if the instance is a VariableScope. Defined here to avoid importing VariableScope and causing a
+     * circular dependency.
+     *
+     * @private
+     * @param {*} obj the object that needs to be validated
+     */
+    isVariableScope = function (obj) {
+        return Boolean(obj) && (_.inSuperChain(obj.constructor, POSTMAN_PROPERTY_NAME,
+            VARIABLE_SCOPE));
+    },
+
+    /**
+     * Converts a variable changeset in the form of a JSON Delta stanza to an object format.
+     *
+     * @private
+     * @param {VariableChangeset~serialized} serialized A serialized variable changeset
+     */
+    deserializeChangeset = function (serialized) {
+        // bail out - empty or invalid
+        if (!(serialized && serialized.length > 1)) {
+            return {};
+        }
+
+        var deserialized = {
+            id: serialized[0],
+            key: serialized[1],
+            type: INSTRUCTION.UNSET // set default changeset type to `unset`
+        };
+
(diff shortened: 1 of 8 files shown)


kamalaknn:
I ran some benchmark tests on these changes,
Iteration x
```
no tracking x 60,505 ops/sec ±7.83% (84 runs sampled)
with tracking x 55,358 ops/sec ±3.49% (83 runs sampled)
with tracking, in compressed mode x 56,838 ops/sec ±1.02% (81 runs sampled)
Fastest is with tracking, in compressed mode,with tracking
```
Iteration y
```
no tracking x 60,763 ops/sec ±9.51% (82 runs sampled)
with tracking x 55,602 ops/sec ±1.00% (88 runs sampled)
with tracking, in compressed mode x 56,535 ops/sec ±0.96% (84 runs sampled)
```
I've added two of the many iterations. The deviation is `8%-10%` for uncompressed tracking and `6%-8%` for compressed tracking. (TIL `obj.foo = 'bar'` is much faster than `array.push('bar')`).

I also ran the memory profile tests on `postman-runtime` with tracking enabled in `postman-sandbox`. 

Collection with a simple `pm.environment.set('foo', 'bar')` in the scripts.
output-small

Collection with a large response `120kb` set as an environment variable. `pm.environment.set('foo', pm.response.text())`.
output

In both the cases, the total average deviation is minuscule.

lib/collection/variable-changeset.js
```diff
@@ -0,0 +1,235 @@
...
+    VARIABLE_SCOPE = 'VariableScope',
+
+    /**
+     * Checks if the instance is a VariableScope. Defined here to avoid importing VariableScope and causing a
+     * circular dependency.
+     *
+     * @private
+     * @param {*} obj the object that needs to be validated
+     */
+    isVariableScope = function (obj) {
```
shamasis: If diff is aware of scope, it should then depend on it and use the `isVariableScope` static function. Since that would be cyclic (though node would handle it owing to singleton nature), we should simply not hard-check VariableScope.

lib/collection/variable-changeset.js
```diff
@@ -0,0 +1,235 @@
...
+        });
+    },
+
+    /**
+     * Applies the changeset on a variable scope
+     *
+     * @param {VariableScope} scope
+     */
+    applyOn: function (scope) {
+        if (!(scope && isVariableScope(scope))) {
```
shamasis: no point checking for isVariableScope. Diff expects an object with set / unset / etc (duck typed. even duck checking would not be needed IMHO)

kamalaknn: It makes sense to not check for `VariableScope`.

kamalaknn:
@shamasis 

Writing down my thought process when I implemented this. Hope that answers the questions.

> Why not distinguish changeset raw/compress based on array/object type

I looked at `raw` and `compressed` as a `mode` to track changesets. The `mode` controls the resolution and storage format for changesets. So your operations like `push` and `pop` would need to be treated differently based on the modes. This is not just storage format related(like `compressed` drops some changesets). So I realized `mode` controls more than storage format. So it did not make sense to detect `mode` from storage format. Also, I wanted the option open to add more modes.

>    Why give changeset it's own property? If we treat it simple array / object and do all manis in VariableScopeIdd, will the memory usage not be lower?

My very first implementation was to keep changesets as plain objects. So I had to implement operations like `serialization`,  `deserialization` and `applying the changeset` outside of the changeset. It made sense to give it it's own class. Then I realized I needed some functionality on `property` like `toJSON` and `auto-id generation`. Then I extended from `Property`.

kamalaknn:
Closed by #666

PR #666, opened 2018-06-26
Added support for mutation tracking in VariableScope

kamalaknn:
* Added `MutationTracker`
* Allow change tracking on `set`, `unset` and `clear` in `VariableScope`

### DIFF
--- a/lib/collection/mutation-tracker.js
+++ b/lib/collection/mutation-tracker.js
@@ -1,0 +1,230 @@
+var _ = require('../util').lodash,
+    PropertyBase = require('./property-base').PropertyBase,
+
+    /**
+     * Primitive mutation types.
+     *
+     * @private
+     * @constant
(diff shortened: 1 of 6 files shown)


lib/collection/mutation-tracker.js
```diff
@@ -0,0 +1,235 @@
...
+
+        return mutationCount;
+    },
+
+    /**
+     * Applies all the recorded mutations on a target object.
+     *
+     * @param {*} target Target to apply mutations. Must implement `applyMutation`.
+     */
+    applyOn: function (target) {
```
kamalaknn: Need to make this async

kamalaknn:
@closes #657

lib/collection/mutation-tracker.js
shamasis: waiting for when we need it.

PR #670, opened 2018-06-28
Remove console warn statement in variable scope

kamalaknn:
Postman Collection SDK executes in multiple environments where sometimes `console` is not available like `postman-sandbox` bootcode before sandbox is completely initialized.

I've removed the runtime warning. The docs still communicate this.

Enabled lint rule to prevent console statements from being added.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -16,7 +16,7 @@
     "no-await-in-loop": "error",
     "no-compare-neg-zero": "error",
     "no-cond-assign": "error",
-    "no-console": "off",
+    "no-console": "error",
     "no-constant-condition": "error",
     "no-control-regex": "error",
     "no-debugger": "error",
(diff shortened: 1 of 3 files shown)


npm/build-docs.js
```diff
@@ -4,6 +4,7 @@
 // ---------------------------------------------------------------------------------------------------------------------
 
 /* eslint-env node, es6 */
+/* eslint no-console: off */
```
kunagpal: All these rules can be placed in `npm/.eslintrc` instead of repeating them across all files.

.eslintrc
```diff
@@ -16,7 +16,7 @@
     "no-await-in-loop": "error",
     "no-compare-neg-zero": "error",
     "no-cond-assign": "error",
-    "no-console": "off",
+    "no-console": "error",
```
shamasis: 👍

PR #673, opened 2018-06-29
JSDoc: Fix `Property~definition` property

### LINKED ISSUE #587: Property constructor uses definition object to check disabled property
In the Property constructor the disabled property is looked for on the definition object which is passed as a parameter. If the object passed as a parameter has an info property this is used as the definition and stored in variable ["src"](https://github.com/postmanlabs/postman-collection/blob/270ed3df11995ea661bc6e1b774cdb432b10ff1d/lib/collection/property.js#L46). However, the disabled property is still [looked for](https://github.com/postmanlabs/postman-collection/blob/270ed3df11995ea661bc6e1b774cdb432b10ff1d/lib/collection/property.js#L87) on the definition object not the src object. 

Is it correct that the object getting passed as the definition parameter can be:
```
{
  info: { id, name}
  disabled
}
```
Which makes the info property not match the Property\~Definition type.

codenirvana:
Fixes #587

### DIFF
--- a/lib/collection/property.js
+++ b/lib/collection/property.js
@@ -14,6 +14,9 @@ var _ = require('../util').lodash,
  * @property {String=} [id]
  * @property {String=} [name]
  * @property {Boolean=} [disabled]
+ * @property {Object=} [info]
+ * @property {String=} [info.id]
+ * @property {String=} [info.name]
  */
 _.inherit((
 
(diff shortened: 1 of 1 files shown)


lib/collection/property.js
```diff
@@ -14,6 +14,9 @@ var _ = require('../util').lodash,
  * @property {String=} [id]
  * @property {String=} [name]
  * @property {Boolean=} [disabled]
+ * @property {Object=} [info]
```
kamalaknn: Can you add jsdoc descriptions for all fields?

PR #681, opened 2018-07-19
Url.parse: Add `variable` property in parsed Url

codenirvana:
- Add `variable` property in parsed Url.
- Fix undefined path variable key name issue from `/:/:/`.

> `variable` property will help App to identify path variables from the raw URL.

### DIFF
--- a/lib/collection/url.js
+++ b/lib/collection/url.js
@@ -243,7 +243,7 @@ _.assign(Url.prototype, /** @lends Url.prototype */ {
             segmentArray = (options && options.unresolved) ? this.path : _.transform(this.path, function (res, segment) {
                 var variable;
 
-                if (_.startsWith(segment, PATH_VARIABLE_IDENTIFIER)) {
+                if (_.startsWith(segment, PATH_VARIABLE_IDENTIFIER) && segment.length > 1) {
                     variable = self.variables.one(segment.slice(1));
                 }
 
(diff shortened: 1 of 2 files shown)


lib/collection/url.js
```diff
@@ -400,6 +400,13 @@ _.assign(Url, /** @lends Url */ {
 
         // extract the hash
         p.hash = _.get(url.match(regexes.extractSearch), GET_1);
+
+        p.variable = _.transform(p.path, function (res, segment) {
+            if (_.startsWith(segment, PATH_VARIABLE_IDENTIFIER) && segment.length > 1) {
```
shamasis: @codenirvana - i know you have explained this to me once... but I keep forgetting the nature of the change added by `segment.length > 1` -> can you document it inline so that in future it is more clear why we did this.

lib/collection/url.js
```diff
@@ -243,7 +243,7 @@ _.assign(Url.prototype, /** @lends Url.prototype */ {
             segmentArray = (options && options.unresolved) ? this.path : _.transform(this.path, function (res, segment) {
                 var variable;
 
-                if (_.startsWith(segment, PATH_VARIABLE_IDENTIFIER)) {
+                if (_.startsWith(segment, PATH_VARIABLE_IDENTIFIER) && segment.length > 1) {
```
shamasis: @codenirvana - i know you have explained this to me once... but I keep forgetting the nature of the change added by segment.length > 1 -> can you document it inline so that in future it is more clear why we did this.

PR #689, opened 2018-08-07
 Add `contentType` in form-param

codenirvana:


lib/collection/form-param.js
```diff
@@ -23,6 +23,7 @@ _.inherit((
         this.value = _.get(options, 'value') || '';
         this.type = _.get(options, 'type');
         this.src = _.get(options, 'src');
+        this.contentType = _.get(options, 'contentType');
```
kunagpal: While this is fine for now, can you add a todo before the first `_.get` call to get rid of all these `_.get` calls?

kunagpal: Also, shouldn't we validate/sanitize the `contentType` value first?

test/unit/form-param.test.js
```diff
@@ -22,4 +22,27 @@ describe('FormParam', function () {
...
+                value: '',
+                src: 'fileSrc',
+                contentType: 'text/csv',
+                type: 'file'
+            });
+        });
+
+        it('should default to undefined if contentType is not provided', function () {
+            var fp = new FormParam({ key: 'foo', value: 'bar' });
+            expect(fp.toJSON()).to.eql({ key: 'foo', value: 'bar' });
```
kunagpal: The assertion should explicitly check for `contentType` being `undefined`.

test/unit/form-param.test.js
```diff
@@ -22,4 +22,27 @@ describe('FormParam', function () {
...
+
+    describe('contentType', function () {
+        it('should parse contentType properly', function () {
+            var fp = new FormParam({
+                key: 'fileName',
+                src: 'fileSrc',
+                contentType: 'text/csv',
+                type: 'file'
+            });
+            expect(fp.toJSON()).to.eql({
```
kunagpal: Newline after `var`.

test/unit/form-param.test.js
```diff
@@ -22,4 +22,27 @@ describe('FormParam', function () {
...
+                key: 'fileName',
+                value: '',
+                src: 'fileSrc',
+                contentType: 'text/csv',
```
kunagpal: The test should only check for the existence of `contentType` in the JSON result.

lib/collection/form-param.js
codenirvana: we should accept any standard/custom `content-type`.

test/unit/form-param.test.js
codenirvana: `.eql` explicitly matches the expected object. There will be no `contentType` key.

test/unit/form-param.test.js
codenirvana: following the convention used for rest of the tests.

PR #701, opened 2018-09-07
Add `disabled` property in PostmanRequestBody

codenirvana:


test/unit/request-body.test.js
```diff
@@ -151,6 +151,15 @@ describe('RequestBody', function () {
             expect(body).to.have.property('mode', 'raw');
             expect(body).to.have.property('raw', 'foo');
         });
+
+        it('should handle disabled property correctly', function () {
```
kunagpal: Also add a test for the reverse case, disabled -> enabled

PR #702, opened 2018-09-09
[DO NOT MERGE] Added `forceInclude` property in PostmanRequestBody

codenirvana:


kunagpal:
Code is fine, why is this needed again?

codenirvana:
@kunagpal To force include request body for methods like `GET`, `HEAD` etc.
Will keep a map of these methods which don't expect the body to be included in the request and this flag is used to alter that behavior.

Here I am considering all custom methods will include request body by default and this flag is not required for them.

kunagpal:
@codenirvana Closing in favour of https://github.com/postmanlabs/postman-collection/pull/707

PR #763, opened 2018-12-27
Fix variable string substitutions

codenirvana:
Reference: https://github.com/postmanlabs/postman-app-support/issues/5682

### DIFF
--- a/lib/superstring/index.js
+++ b/lib/superstring/index.js
@@ -51,7 +51,7 @@ _.assign(SuperString.prototype, /** @lends SuperString.prototype */ {
         }));
 
         this.replacements = replacements; // store the last replacements
-        this.substitutions += Number(replacements); // if any replacement is done, count that some substitution was made
+        replacements && (this.substitutions += 1); // if any replacement is done, count that some substitution was made
         return this;
     },
 
--- a/test/unit/property.test.js
+++ b/test/unit/property.test.js
@@ -215,8 +215,66 @@ describe('Property', function () {
     });
 
     describe('.replaceSubstitutions', function () {
+        /**
+         * Generates a object with n variables
+         *
+         * @example
+         * getVariables(2)
+         * {
+         *    '0': '',
+         *    '1': ''
+         * }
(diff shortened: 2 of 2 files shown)


test/unit/property.test.js
```diff
@@ -215,9 +215,67 @@ describe('Property', function () {
...
+            return obj;
+        }
+
         it('should bail out if a non-string argument is passed', function () {
             expect(Property.replaceSubstitutions(['random'])).to.eql(['random']);
         });
+
+        it('should resolve poly chained variable 19 times', function () {
+            // eslint-disable-next-line max-len
+            var str = '{{20{{19{{18{{17{{16{{15{{14{{13{{12{{11{{10{{9{{8{{7{{6{{5{{4{{3{{2{{1{{0}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}',
```
kunagpal: This will have to be updated if the limit ever changes.

lib/superstring/index.js
```diff
@@ -51,7 +51,7 @@ _.assign(SuperString.prototype, /** @lends SuperString.prototype */ {
         }));
 
         this.replacements = replacements; // store the last replacements
-        this.substitutions += Number(replacements); // if any replacement is done, count that some substitution was made
+        replacements && (this.substitutions += 1); // if any replacement is done, count that some substitution was made
```
kunagpal: Is the substitutions property being used elsewhere?

test/unit/property.test.js
codenirvana: Yes, any alternative?

lib/superstring/index.js
codenirvana: No.

test/unit/property.test.js
kunagpal: Have a function that accepts a number argument and builds this string for you.

>

codenirvana: done

PR #768, opened 2019-01-02
Add option to support duplicate headers in getHeaders method

codenirvana:


test/unit/request.test.js
```diff
@@ -282,6 +282,46 @@ describe('Request', function () {
...
+                        key: 'name',
+                        value: 'beta'
+                    }, {
+                        key: 'name',
+                        value: 'gamma',
+                        disabled: true
+                    }]
+                },
+                request = new Request(rawRequest);
+            expect(request.getHeaders({ enabled: true, multiValue: true })).to.eql({
```
kunagpal: Also add a test that checks for this being stringified correctly.

codenirvana: This won't affect the `toString` flow, multiple `key-value` header pairs will be returned irrespective of the duplicates.

lib/collection/request.js
```diff
@@ -155,14 +155,15 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
      * @param {Object=} options
      * @param {Boolean} options.ignoreCase When set to "true", will ensure that all the header keys are lower case.
      * @param {Boolean} options.enabled Only get the enabled headers
-     *
+     * @param {Boolean} options.multiValue When set to "true", duplicate header values will be stored in an array
```
shamasis: @codenirvana - does thing break, if we allow multi value by default?

codenirvana: Yes, `getHeaders` method is used for `aws4` auth which might fail if duplicates are present.
Not sure about the current state, possibly having issues with current implementation as well 🤔

PR #769, opened 2019-01-02
Avoid resolving disabled variables

codenirvana:


lib/superstring/index.js
```diff
@@ -105,7 +106,10 @@ _.assign(Substitutor.prototype, /** @lends Substitutor.prototype */ {
 
             // in case the object is a postman variable list, we give special attention
             if (obj.constructor._postman_propertyName === 'VariableList') {
-                if (obj.has(key)) { return obj.one(key); }
+                value = obj.one(key);
+                if (value && !value.disabled) {
```
shamasis: The `.has` check seemed useful to keep.

PR #770, opened 2019-01-02
Add options to listeners & listenersOwn methods

codenirvana:
- Add option to filter disabled events

PR #771, opened 2019-01-02
Optimized listeners and listenersOwn methods

codenirvana:
Resolves comments from #770


lib/collection/event-list.js
```diff
@@ -63,10 +61,14 @@ _.assign(EventList.prototype, /** @lends EventList.prototype */ {
      * @returns {Array<Event>}
      */
     listenersOwn: function (name, options) {
-        !options && (options = {});
+        if (options && options.excludeDisabled) {
+            return this.filter(function (event) {
```
kunagpal: Repetition with `this.filter`

shamasis: Elaborate?

PR #772, opened 2019-01-03
Inherit propertySanitizeKeys option in PropertyList

codenirvana:
Extends #430

Using this `PropertyList` will inherit `propertySanitizeKeys` option which will be useful in cases like`HeaderList` or any other list where falsy keys are not allowed.

Property `_postman_listSanitizeKeys` used in following method:
- `toObject`


codenirvana:
**DO NOT DELETE** : `feature/sanitizeKeys`

PR #780, opened 2019-01-09
Reorganised URL parameters to always honour `disabled` status

codenirvana:


lib/collection/url.js
```diff
@@ -238,81 +238,116 @@ _.assign(Url.prototype, /** @lends Url.prototype */ {
     /**
      * Returns the request path, with a leading '/'.
      *
-     * @param {Object=} options
-     * @param {Object} options.unresolved If set to true, path variables will not be processed
+     * @param {?Booleamn=} [unresolved=false]
```
kunagpal: `Boolean`

lib/collection/url.js
```diff
@@ -238,81 +238,116 @@ _.assign(Url.prototype, /** @lends Url.prototype */ {
...
+
+        segments = _.transform(this.path, function (res, segment) {
+            var variable;
+
+            // check if the segment has path variable prefix followed by the variable name.
+            if (_.startsWith(segment, PATH_VARIABLE_IDENTIFIER) && segment !== PATH_VARIABLE_IDENTIFIER) {
+                variable = self.variables.one(segment.slice(1)); // remove path variable prefix.
+            }
+
+            variable = variable && variable.valueOf && variable.valueOf();
```
kunagpal: Shift the assignment to the if block. `variable` would be undefined in case the `if` block was not hit.

lib/collection/url.js
```diff
@@ -238,81 +238,116 @@ _.assign(Url.prototype, /** @lends Url.prototype */ {
...
+     *
+     * @note Deprecated variant of this function is as follows:
+     * - param {?Object} [options={}]
+     * - param {?Boolean} options.encode - Enables URL encoding when processing the query string.
+     * - param {?Boolean} options.ignoreDisabled - Prevents disabled query parameters from showing up in the unparsed
      */
-    getQueryString: function (options) {
-        if (this.query.count()) {
-            return QueryParam.unparse(this.query.all(), options);
+    getQueryString: function (encoded) {
```
kunagpal: Present tense.

PR #781, opened 2019-01-10
Added Request~size

codenirvana:


kunagpal:
A few concerns:
1. `_size` instead of just `size`.
2. APIs to set and get size (`set/getSize` vs overloaded `.size`?)

lib/collection/request.js
```diff
@@ -107,7 +107,13 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
              * @type {Certificate|undefined}
              */
-            certificate: options.certificate && new Certificate(options.certificate)
+            certificate: options.certificate && new Certificate(options.certificate),
+
+            /**
+             * @private
+             * @type {Object}
+             */
+            size: options && options.size
```
shamasis: Please refer to `request.size()` as a function. Compute size when called. The non-computable size, IMHO, is body size and that should be returned by adding a .size property in request-body which in turn is used inside request object.

Places where we know the size only after reading, etc - can be stored as a fallback in a private property of the body.

lib/collection/request.js
```diff
@@ -9,7 +10,66 @@ var _ = require('../util').lodash,
...
+     * @const
+     * @type {String}
+     */
+    CONNECTION_KEEP_ALIVE = 'Connection: keep-alive',
+
+    /**
+     * @private
+     * @type {Boolean}
+     */
+    supportsBuffer = (typeof Buffer !== undefined) && _.isFunction(Buffer.byteLength);
```
kunagpal: 👍

lib/collection/request.js
```diff
@@ -107,13 +167,7 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
              * @type {Certificate|undefined}
              */
-            certificate: options.certificate && new Certificate(options.certificate),
-
-            /**
-             * @private
-             * @type {Object}
-             */
-            size: options && options.size
+            certificate: options.certificate && new Certificate(options.certificate)
```
kunagpal: Future scope: The constructor should handle all such truthy checks.

lib/collection/request.js
```diff
@@ -257,6 +311,46 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
+            sizeInfo = {
+                body: 0,
+                header: 0,
+                total: 0
+            };
+
+        // if 'Content-Length' header is present, we take body as declared by
+        // the client(postman-request or user-defined). else we need to compute the same.
+        if (contentLength && util.isNumeric(contentLength)) {
+            sizeInfo.body = _.parseInt(contentLength, 10);
```
kunagpal: Why `_.parseInt`?

lib/collection/request.js
```diff
@@ -257,6 +311,46 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
+        }
+        // otherwise, if body is defined, we calculate the length of the body
+        else if (!_.isNil(this.body)) {
+            sizeInfo.body = supportsBuffer ? Buffer.byteLength(this.body.toString()) : this.body.toString().length;
```
kunagpal: Store the result of `this.body.toString()` and reuse it.

lib/collection/request.js
```diff
@@ -251,6 +311,48 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
+            sizeInfo = {
+                body: 0,
+                header: 0,
+                total: 0
+            };
+
+        // if 'Content-Length' header is present, we take body as declared by
+        // the client(postman-request or user-defined). else we need to compute the same.
+        if (contentLength && util.isNumeric(contentLength)) {
+            sizeInfo.body = parseInt(contentLength);
```
kunagpal: Missing radix.

test/unit/request.test.js
```diff
@@ -565,6 +565,54 @@ describe('Request', function () {
         });
     });
 
+    describe('.size', function () {
```
kunagpal: Add a comment explaining how the default values for empty requests are arrived at.

lib/collection/request.js
```diff
@@ -251,6 +311,48 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
+    /**
+     * Get the request size by computing the headers and body or using the
+     * actual content length header once the request is sent.
+     *
+     * @returns {Object}
+     */
+    size: function () {
+        var contentLength = this.headers.get(CONTENT_LENGTH),
+            requestTarget = this.url.getPathWithQuery(),
+            bodyString = _.isFunction(this.body && this.body.toString) && this.body.toString(),
```
shamasis: do toString only if content length is absent

lib/collection/request.js
```diff
@@ -251,6 +311,50 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
+     * @returns {Object}
+     */
+    size: function () {
+        var contentLength = this.headers.get(CONTENT_LENGTH),
+            requestTarget = this.url.getPathWithQuery(),
+            bodyString,
+            sizeInfo = {
+                body: 0,
+                header: 0,
+                total: 0
```
shamasis: Need to add the “source” section to indicate whether this is from content length or computer.

lib/collection/request.js
```diff
@@ -251,6 +311,50 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
+            };
+
+        // if 'Content-Length' header is present, we take body as declared by
+        // the client(postman-request or user-defined). else we need to compute the same.
+        if (contentLength && util.isNumeric(contentLength)) {
+            sizeInfo.body = parseInt(contentLength, 10);
+        }
+        // otherwise, if body is defined, we calculate the length of the body
+        else if (this.body) {
+            // @note body.toString() returns E for formdata or file mode
```
shamasis: Need to attempt something for form data and file mode + append message in “source” property that computation is not accurate.

codenirvana: Added to my ToDo. Also, we should consider setting our own form-data boundary to better control this?

shamasis: agreed regarding form data boundary.

lib/collection/request.js
```diff
@@ -251,6 +324,52 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
+        }
+
+        // https://tools.ietf.org/html/rfc7230#section-3.1.1
+        // request-line   = method SP request-target SP HTTP-version CRLF
+        sizeInfo.header = (this.method + SP + requestTarget + SP + HTTP_X_X + CRLF).length + this.headers.contentSize();
+
+        // account for connection header which will be added by Node's HTTP Agent
+        // since its not added back via Request~upsertHeader.
+        // @note requester.keepAlive is true by default in postman-runtime
+        if (!this.headers.has('connection')) {
```
kunagpal: Top level const.

PR #783, opened 2019-01-14
Added support for Array and Object variable type

codenirvana:


lib/collection/variable.js
```diff
@@ -280,6 +282,82 @@ _.assign(Variable, /** @lends Variable */ {
...
 
+        /**
+         * A "array" type value stores Array data format
+         */
+        'array': {
+            /**
+             * @param {Array} val
+             * @returns {String}
+             */
+            in: function (val) {
```
kunagpal: Don't lose data on imports, sanitize on export

PR #785, opened 2019-01-18
 Fix Request~upsertHeader on multi-value updates

codenirvana:


lib/collection/request.js
```diff
@@ -223,12 +223,41 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
+        // it's better to remove the existing header and and insert new records.
+        // since it's difficult to correctly update the members and references of
+        // a HeaderList due to multiValue and case insensitive names.
+        // side effects:
+        //   1. loses header index
+        //   2. probably expensive (benchmark this vs finding individual reference)
+        // @todo FIX IT ¯\_(ツ)_/¯
+        if (referenceId && (Array.isArray(self.headers.reference[referenceId]) || multiValue)) {
+            self.removeHeader(header.key);
+            existing = false; // existing header is removed
```
kunagpal: Remove this assignment and the `!existing` check in the block below. The next `if` block below it will become an else in that case.

lib/collection/request.js
```diff
@@ -223,12 +223,41 @@ _.assign(Request.prototype, /** @lends Request.prototype */ {
...
+            existing = this.headers.find({ key: header.key }), // first match
+            referenceId = existing && (self.headers._postman_listIndexCaseInsensitive ?
+                existing.key.toLowerCase() : existing.key);
+
+        // if there's an existing header with multiValue or the given input have multiValue,
+        // it's better to remove the existing header and and insert new records.
+        // since it's difficult to correctly update the members and references of
+        // a HeaderList due to multiValue and case insensitive names.
+        // side effects:
+        //   1. loses header index
```
kunagpal: This is a problem.

PR #787, opened 2019-01-21
Handle disabled headers on unparse

codenirvana:


lib/collection/header.js
```diff
@@ -226,7 +226,14 @@ _.assign(Header, /** @lends Header */ {
             return E;
         }
 
-        return headers.map(Header.unparseSingle).join(separator ? separator : '\n');
+        return (Array.isArray(headers) ? headers : headers.members)
```
kunagpal: Add a reduce helper to `PropertyList` and utilize that instead.

lib/collection/header.js
```diff
@@ -226,7 +226,14 @@ _.assign(Header, /** @lends Header */ {
...
+            .reduce(function (acc, header) {
+                var unparsed = Header.unparseSingle(header);
+                unparsed && acc.push(unparsed);
+
+                return acc;
+            }, [])
+            .join(separator ? separator : '\n');
```
kunagpal: Check if a `EOL`/`NEWLINE`/`LF` shared const exists above.

lib/collection/header.js
```diff
@@ -236,7 +243,8 @@ _.assign(Header, /** @lends Header */ {
      * @returns {String}
      */
     unparseSingle: function (header) {
-        if (!_.isObject(header)) { return E; }
+        if (!(_.isObject(header) && header.key && !header.disabled)) { return E; }
```
kunagpal: Add a comment questioning this approach for falsy keys

lib/collection/header.js
```diff
@@ -226,7 +227,14 @@ _.assign(Header, /** @lends Header */ {
             return E;
         }
 
-        return headers.map(Header.unparseSingle).join(separator ? separator : '\n');
+        return headers.reduce(function (acc, header) {
+            // unparse header only if it has a valid key and is not disabled
+            if (header && header.key && !header.disabled) {
```
shamasis: Is blank key a problem? I know … naive question.

codenirvana: Yes, it throws `ERR_INVALID_HTTP_TOKEN`

codenirvana: @shamasis @kunagpal Issue with Node's HTTP Parser. Works fine with curl and non-NodeJS servers.
I think the best place to handle this specific behavior is in `HeaderList~contentSize`.

shamasis: LGTM otherwise,

PR #792, opened 2019-01-28
Fix headers size calculation

codenirvana:
- Fix HTTP Message calculation
- Handle disabled and falsy key headers 

Closes #787 

```
   [0]: join x 60,090,816 ops/sec ±0.18% (94 runs sampled)
   [0]: length + join x 181,755,451 ops/sec ±0.10% (96 runs sampled)
[1000]: join x 49,984 ops/sec ±0.94% (92 runs sampled)
[1000]: length + join x 50,851 ops/sec ±0.23% (92 runs sampled)
Fastest is [0]: length + join
```

### DIFF
--- a/lib/collection/header-list.js
+++ b/lib/collection/header-list.js
@@ -30,21 +30,15 @@ _.assign(HeaderList.prototype, /** @lends HeaderList.prototype */ {
      * @returns {Number}
      */
     contentSize: function () {
-        var raw = E,
+        var raw = this.reduce(function (acc, header) {
+            // unparse header only if it has a valid key and is not disabled
+            if (header && !header.disabled) {
+                // *( header-field CRLF )
+                acc += Header.unparseSingle(header) + CRLF;
+            }
 
-            headers = this.reduce(function (acc, header) {
-                // unparse header only if it has a valid key and is not disabled
-                if (header && !header.disabled) {
-                    acc.push(Header.unparseSingle(header));
-                }
-
-                return acc;
-            }, []);
-
-        if (headers.length) {
-            // *( header-field CRLF )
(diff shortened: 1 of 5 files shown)


lib/collection/header-list.js
```diff
@@ -30,11 +30,22 @@ _.assign(HeaderList.prototype, /** @lends HeaderList.prototype */ {
      * @returns {Number}
      */
     contentSize: function () {
-        if (!this.count()) { return 0; }
+        var raw = E,
+
+            headers = this.reduce(function (acc, header) {
+                // unparse header only if it has a valid key and is not disabled
+                if (header && header.key && !header.disabled) {
```
shamasis: Let’s not extract header.key here. I know that size will be a bit off - but for really edge cases and we don’t want to bleed the node issue all over our codebase.

lib/collection/header-list.js
```diff
@@ -30,11 +30,22 @@ _.assign(HeaderList.prototype, /** @lends HeaderList.prototype */ {
      * @returns {Number}
      */
     contentSize: function () {
-        if (!this.count()) { return 0; }
+        var raw = E,
+
+            headers = this.reduce(function (acc, header) {
```
kunagpal: Why can't you append to the string `raw` directly?

codenirvana: It will be like this:
```javascript
raw = this.reduce(..).join(CRLF);
raw !== E && (raw += CRLF);
```

unparsing headers separately looks cleaner, but the alternative I can think of is one shared above.

kunagpal:
@codenirvana Need benchmarks too.

PR #799, opened 2019-02-08
Support variables with # in query parameters - Fixes Runtime 5779

ankitsiam:
By default query is chopped at # but this shouldn't be the case for variables, atleast not until they are resolved thus # should be allowed in variable names. Fixes https://github.com/postmanlabs/postman-app-support/issues/5779

### DIFF
--- a/lib/collection/url.js
+++ b/lib/collection/url.js
@@ -34,7 +34,7 @@ var _ = require('../util').lodash,
         extractPort: /:([^:]*)$/,
         extractPath: /.*?(?=\?|#|$)/,
         trimPath: /^\/((.+))$/,
-        extractQuery: /^\?((([^#]*[=&%()==@[\]?:/.!-]))([^#{]*)|(\{\{([^]*?)}}*))*/g,
+        extractQuery: /^\?((([^#]*[=&%()@[\]?:/.!-]))|(\{{2}([^]*?)*\}{2})|(\{*[^#]))*/g,
         extractSearch: /#(.+)$/,
         splitDomain: /\.(?![^{]*\}{2})/g
     },
--- a/lib/collection/url.js
+++ b/lib/collection/url.js
@@ -34,7 +34,7 @@ var _ = require('../util').lodash,
         extractPort: /:([^:]*)$/,
         extractPath: /.*?(?=\?|#|$)/,
         trimPath: /^\/((.+))$/,
-        extractQuery: /^\?([^#]+)/,
+        extractQuery: /^\?((([^#]*[=&%()==@[\]?:/.!-]))([^#{]*)|(\{\{([^]*?)}}*))*/g,
         extractSearch: /#(.+)$/,
         splitDomain: /\.(?![^{]*\}{2})/g
     },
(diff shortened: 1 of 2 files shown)


kunagpal:
@ankitsiam Needs before and after benchmarks for variables with and without `#` in them.

test/unit/url.test.js
```diff
@@ -951,6 +951,47 @@ describe('Url', function () {
             var url = new Url('http://127.0.0.1/hello/world/?query=param&query2=param2#test-api');
             expect(url.query.toObject()).to.eql({ query: 'param', query2: 'param2' });
         });
+
+        describe('with # in them', function () {
+            it('should ignore # if used for variable names', function () {
```
kunagpal: Change the test title to read: `should not truncate variables at with # in their names`, the current content is misleading.

lib/collection/url.js
```diff
@@ -443,8 +443,9 @@ _.assign(Url, /** @lends Url */ {
         }
 
         // extract the query string
-        p.query = _.get(url.match(regexes.extractQuery), GET_1);
-        _.isString(p.query) && ((url = url.substr(p.query.length + 1)), (p.query = QueryParam.parse(p.query)));
+        p.query = _.get(url.match(regexes.extractQuery), GET_0);
+        _.isString(p.query) && (p.query = p.query.substring(p.query.indexOf('?') + 1)) &&
```
kunagpal: Love the way this is so easy on the eye 😝

lib/collection/url.js
```diff
@@ -34,7 +34,7 @@ var _ = require('../util').lodash,
         extractPort: /:([^:]*)$/,
         extractPath: /.*?(?=\?|#|$)/,
         trimPath: /^\/((.+))$/,
-        extractQuery: /^\?([^#]+)/,
+        extractQuery: /^\?((([^#]*[=&%()==@[\]?:/.!-]))([^#{]*)|(\{\{([^]*?)}}*))*/g,
```
codenirvana: This might break lots of things 😕
e.g `?q={{` is not matching.

ankitsiam: Isn't that expected behaviour ? If you don't have a closing }} it's not a valid syntax hence variable won't be resolved ?

codenirvana: No, basic rule: anything in between `?` & `#` is the query string.
Allowing `#` in the variable name(inside query string) is what we have to support.

ankitsiam: As per rfc 3986, If data for a URI component would conflict with a reserved character's purpose as a delimiter, then the conflicting data must be percent-encoded before the URI is formed.

And it defines following as reserved characters

reserved = gen-delims / sub-delims

gen-delims = ":" / "/" / "?" / "#" / "[" / "]" / "@"

sub-delims = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="

shamasis:
Closing based on https://github.com/postmanlabs/postman-app-support/issues/5779#issuecomment-465018548

PR #801, opened 2019-02-11
Support variables with # in query parameters - Fixes Runtime 5779

ankitsiam:
By default query is chopped at # but this shouldn't be the case for variables, atleast not until they are resolved thus # should be allowed in variable names. Fixes [postmanlabs/postman-app-support#5779](https://github.com/postmanlabs/postman-app-support/issues/5779)

Alternative methods using regex : https://github.com/postmanlabs/postman-collection/pull/799


shamasis:
Closing based on https://github.com/postmanlabs/postman-app-support/issues/5779#issuecomment-465018548

PR #825, opened 2019-03-06
Added support for proxy authentication

codenirvana:
- Added `auth` property in `ProxyConfig`
- Only basic auth method is supported
- Updated `ProxyConfig#getProxyUrl` to add auth in URL

### DIFF
--- a/lib/collection/proxy-config.js
+++ b/lib/collection/proxy-config.js
@@ -11,6 +11,7 @@ var _ = require('../util').lodash,
     DEFAULT_PORT = 8080,
     PROTOCOL_HOST_SEPARATOR = '://',
     MATCH_ALL_HOST_AND_PATH = '*/*',
+    AUTH_CREDENTIALS_SEPARATOR = '@',
     DEFAULT_PROTOCOL = 'http',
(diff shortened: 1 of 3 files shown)


lib/collection/proxy-config.js
```diff
@@ -181,7 +193,12 @@ _.assign(ProxyConfig.prototype, /** @lends ProxyConfig.prototype */ {
      * @returns {String}
      */
     getProxyUrl: function () {
-        return DEFAULT_PROTOCOL + PROTOCOL_HOST_SEPARATOR + this.host + COLON + this.port;
+        var proxyAuth = '',
```
shamasis: E

lib/collection/proxy-config.js
```diff
@@ -181,7 +193,12 @@ _.assign(ProxyConfig.prototype, /** @lends ProxyConfig.prototype */ {
...
+            auth = this.auth && this.auth.type === BASIC_AUTH_METHOD && this.auth.current();
```
shamasis: neatly put this in an if block below and do the assignment.

- single variable for auth string and call it `auth`

PR #843, opened 2019-03-29
Added RequestBody mode `graphql`

codenirvana:
- [x] Add tests


lib/collection/request-body.js
```diff
@@ -68,9 +78,12 @@ _.assign(RequestBody.prototype, /** @lends RequestBody.prototype */ {
         // If mode is urlencoded but options does not provide any content, set it to an empty property list
         (mode === RequestBody.MODES.urlencoded && !urlencoded) && (urlencoded = new PropertyList(QueryParam, this, []));
 
-        // If mode is urlencoded but options does not provide any content, set it to an empty property list
+        // If mode is formdata but options does not provide any content, set it to an empty property list
         (mode === RequestBody.MODES.formdata && !formdata) && (formdata = new PropertyList(FormParam, this, []));
 
+        // If mode is graphql but options does not provide any content, set empty query
+        (mode === RequestBody.MODES.graphql && !graphql) && (graphql = { query: EMPTY });
```
shamasis: Why empty query? I mean, it would be better to write down here or somewhere relevant, why are we:

1. not sending blank body or a body having `{}`
2. Why did we choose query key and not other keys in default

lib/collection/request-body.js
```diff
@@ -151,12 +173,22 @@ _.assign(RequestBody.prototype, /** @lends RequestBody.prototype */ {
...
+        // Handle file mode
         if (mode === RequestBody.MODES.file) {
-            // Handle file mode.
-            return !(data && (data.src || data.content));
+            return !(data.src || data.content);
         }
 
-        return !data || !(_.isString(data) ? data.length : (_.isFunction(data.count) && data.count()));
+        // Handle graphql mode, send empty query anyways
+        if (mode === RequestBody.MODES.graphql) {
```
shamasis: Well... I think we need to check if the body type stores in object format and it has no keys ... like lodash isEmpty of some sort and not hard-code graphql here. This keeps things still manageable when we make future extensions and also keeps in line with present.

lib/collection/request-body.js
```diff
@@ -4,9 +4,9 @@ var _ = require('../util').lodash,
     QueryParam = require('./query-param').QueryParam,
     FormParam = require('./form-param').FormParam,
 
-    RequestBody,
+    EMPTY = '',
```
shamasis: `E`

PR #883, opened 2019-06-11
Fixed a bug where VariableScope.get() method falsely returned undefined

vikiCoder:
- Fixed a bug where `VariableScope.get()` method returned `undefined` if a variable with a duplicate key is present in both, current scope and any parent scopes, but it is disabled in current scope and enabled in parent scope.
- Corrected behavior: It should ignore disabled variable of current scope and return enabled variable from parent scope.

Example:
```javascript
var scope = new VariableScope([
    { key: 'alpha', value: 'foo', disabled: true }
]);

scope.addLayer(new VariableList({}, [
    { key: 'alpha', value: 'foo_layer_1' }
]));

console.log(scope.get('alpha'));
/* Output:
* => undefined // before this fix
* => foo_layer_1 // after this fix
*/
```

### DIFF
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -1,11 +1,10 @@
 master:
-  fixed bugs:
-    - >-
-      GH-883 Fixed a bug where `VariableScope~get` method falsely returned `undefined` when variable
-      with same name is present in both, current and parent scopes, but it is disabled in current scope
-      and enabled in parent scope.
   new features:
     - GH-882 Updated `VariableScope~has` method to search in all the scopes
+  fixed bugs:
+    - >-
+      GH-883 Fixed a bug where `VariableScope~get` method doesn't search
+      all the scopes if the variable is disabled in local scope
 
 3.4.9:
   date: 2019-06-07
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -1,5 +1,5 @@
 master:
-  new features:
+  fixed bugs:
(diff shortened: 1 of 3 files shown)


CHANGELOG.yaml
```diff
@@ -1,3 +1,7 @@
+master:
+  fixed bugs:
+    - GH-883 Fixed a bug where `VariableScope~get` method falsely returned `undefined`
```
codenirvana: Also, specify `why` or `when`. And keep it concise.

test/unit/variable-scope.test.js
```diff
@@ -290,67 +290,118 @@ describe('VariableScope', function () {
...
+                    ]);
+
+                    scope.addLayer(new VariableList({}, [
+                        { key: 'alpha', value: 'foo_layer_1' }
+                    ]));
+
+                    expect(scope.get('alpha')).to.equal('foo');
+                });
+
+                it('should pick from first layer with enabled value in case of duplicates', function () {
```
codenirvana: assertion title can be improved.

lib/collection/variable-scope.js
```diff
@@ -214,7 +214,7 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
             ii;
 
         // if a variable does not exist in local scope, we search all the layers and return the first occurrence.
-        if (!(variable || !this._layers)) {
+        if ((!variable || variable.disabled) && this._layers) {
```
shamasis:
```suggestion
        if ((!variable || (variable.disabled === true)) && this._layers) {
```
for protection?

PR #893, opened 2019-06-25
Add a new method `getAll` to PropertyList

coditva:
This method gets all the values of an item in the list. If multiple values are not allowed, only the last item is returned. The return type is always an array.

Ref: https://github.com/postmanlabs/postman-app-support/issues/6143

### DIFF
--- a/lib/collection/property-list.js
+++ b/lib/collection/property-list.js
@@ -389,6 +389,30 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
     },
 
     /**
+     * Get all the values of an item in the list. If multiple values are not allowed, only the last item is returned.
+     *
+     * @param {String|Function} key
+     * @returns {Array<PropertyList.Type|*>}
+     */
+    getAll: function (key) {
+        var member,
+            val = [];
+
+        member = this.reference[this._postman_listIndexCaseInsensitive ? String(key).toLowerCase() : key];
+        if (!member) { return; } // eslint-disable-line getter-return
+
+        if (this._postman_listAllowsMultipleValues && Array.isArray(member)) {
+            _.forEach(member, function (obj) {
+                val.push(obj.valueOf());
+            });
+            return val;
+        }
+
+        val.push(member.valueOf());
+        return val;
+    },
+
+    /**
      * Iterate on each item of this list.
      *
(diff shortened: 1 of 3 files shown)


lib/collection/property-list.js
```diff
@@ -388,6 +388,30 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
         return member.valueOf();
     },
 
+    /**
+     * Get all the values of an item in the list. If multiple values are not allowed, only the last item is returned.
+     *
+     * @param {String|Function} key
```
codenirvana:
```suggestion
     * @param {String} key
```

lib/collection/property-list.js
```diff
@@ -388,6 +388,30 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
+     * @returns {Array<PropertyList.Type|*>}
```
codenirvana:
```suggestion
     * @returns {Array<PropertyList.Type>|*}
```

lib/collection/property-list.js
```diff
@@ -388,6 +388,30 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
+     */
+    getAll: function (key) {
+        var member,
+            val = [];
+
+        member = this.reference[this._postman_listIndexCaseInsensitive ? String(key).toLowerCase() : key];
+        if (!member) { return; } // eslint-disable-line getter-return
```
codenirvana:
```suggestion
        if (!member) { return; }
```

lib/collection/property-list.js
```diff
@@ -388,6 +388,30 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
+
+        if (this._postman_listAllowsMultipleValues && Array.isArray(member)) {
```
codenirvana: consider `_postman_listAllowsMultipleValues` false.

lib/collection/property-list.js
```diff
@@ -388,6 +388,30 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
...
+            _.forEach(member, function (obj) {
```
codenirvana: _.clone

PR #897, opened 2019-07-01
Remove irrelevant and incomplete faker generators

coditva:
Removed:
- `address.cityPrefix`
- `address.citySuffix`
- `address.streetPrefix`
- `address.streetSuffix`
- `address.secondaryAddress`
- `address.county`
- `commerce.price`
- `date.between`
- `name.title`
- `phone.phoneNumberFormat`
- `phone.phoneFormats`
- `random.arrayElement`
- `random.objectElement`
- `system.directoryPath` (return `undefined`)
- `system.filePath` (return `undefined`)

### DIFF
--- a/test/unit/superstring.test.js
+++ b/test/unit/superstring.test.js
@@ -52,5 +52,12 @@ describe('String utilities', function () {
                 generatorSet[extension] = name;
             });
         });
+
+        it('should not include generators that are incomplete', function () {
+            _.forOwn(fakermap, function (extension) {
+                var generator = _.get(faker, extension);
+                expect(generator()).to.not.be.undefined;
+            });
+        });
     });
 });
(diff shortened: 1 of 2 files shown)


test/unit/superstring.test.js
```diff
@@ -52,5 +52,12 @@ describe('String utilities', function () {
                 generatorSet[extension] = name;
             });
         });
+
+        it('should not include generators that are incomplete', function () {
```
codenirvana: How it's different from the first assertion `should contain required generators`?
And how incomplete generators like `system.directoryPath` spilled through it?

coditva: `FakerJS` has a `directoryPath` generator which is marked TODO and returns nothing. Because it _is_ a function, it does not get caught in `.to.be.a('function')`.

coditva:
Rebased on `develop`

PR #904, opened 2019-07-04
Add descriptions for dynamic variables

coditva:
Change the `dynamicGenerators` object in `lib/superstring/dynamic-variables.js` to:
```js
    {
      // before:
      // guid: function () {
      //   return uuid.v4();
      // }

      // proposed:
      $guid: {
        name: '$guid',
        description: 'Generates a v4 style guid',
        generator: function () {
            return uuid.v4();
        }
      },
      $randomCity: {
        name: '$randomCity',
        description: 'Generates a random city name',
        generator: faker.address.city
      }
    }
```

This allows us to add descriptions for each dynamic variable.


lib/superstring/index.js
```diff
@@ -222,9 +221,9 @@ _.assign(Substitutor, /** @lends Substitutor */ {
 });
 
 // @todo make the default variales of SuperString extensible and do this anywhere else but here
-_.forOwn(dynamicGenerators, function (generator, name) {
-    if (_.isFunction(generator)) {
-        Substitutor.DEFAULT_VARS[DYN_VAR_BASE + name] = generator;
+_.forOwn(dynamicVariables, function (variable, name) {
+    if (_.isFunction(variable.generator)) {
+        Substitutor.DEFAULT_VARS[name] = variable.generator;
```
codenirvana: Also set `DEFAULT_VARS: {}`.

lib/superstring/index.js
```diff
@@ -222,9 +221,9 @@ _.assign(Substitutor, /** @lends Substitutor */ {
 });
 
 // @todo make the default variales of SuperString extensible and do this anywhere else but here
-_.forOwn(dynamicGenerators, function (generator, name) {
-    if (_.isFunction(generator)) {
-        Substitutor.DEFAULT_VARS[DYN_VAR_BASE + name] = generator;
+_.forOwn(dynamicVariables, function (variable, name) {
+    if (_.isFunction(variable.generator)) {
```
codenirvana:
```suggestion
    if (variable && _.isFunction(variable.generator)) {
```

test/unit/dynamic-variables.test.js
```diff
@@ -0,0 +1,111 @@
...
+describe('Dynamic variable', function () {
+    it('should have all attributes', function () {
+        _.forOwn(dynamicVariables, function (dynamicVariable) {
+            expect(dynamicVariable).to.be.an('object');
+            expect(dynamicVariable.description).to.be.a('string');
+            expect(dynamicVariable.generator).to.be.a('function');
+        });
+    });
+
+    it('should not be a duplicate', function () {
```
codenirvana: This test is no longer required because we moved away from the faker-map strategy and now the value is stored as an object.

test/unit/dynamic-variables.test.js
```diff
@@ -0,0 +1,111 @@
+var _ = require('lodash'),
+    dynamicVariables = require('../../lib/superstring/dynamic-variables'),
+    expect = require('chai').expect;
+
+describe('Dynamic variable', function () {
+    it('should have all attributes', function () {
+        _.forOwn(dynamicVariables, function (dynamicVariable) {
```
codenirvana: Add an assertion above which checks for the type of `dynamicVariables`.

test/unit/dynamic-variables.test.js
```diff
@@ -0,0 +1,111 @@
...
+        });
+    });
+
+
+    describe('generator', function () {
+        it('should return random data', function () {
+            _.forOwn(dynamicVariables, function (variable) {
+                expect(variable.generator()).to.not.be.undefined;
+            });
+        });
```
codenirvana:
```suggestion
        });
        
        it('$randomInt: should return a random number', function () {
            expect(dynamicVariables.$randomInt.generator()).to.be.within(0, 1000);
        });
```

test/unit/dynamic-variables.test.js
```diff
@@ -0,0 +1,111 @@
...
+    });
+
+    describe('$randomInt', function () {
```
codenirvana: move all the `$*` describe blocks inside `generator` as an `it` block (an example is given above).

lib/superstring/dynamic-variables.js
```diff
@@ -0,0 +1,616 @@
...
+        '/var',
+        '/var/log',
+        '/var/mail',
+        '/var/spool',
+        '/var/tmp',
+        '/var/yp'
+    ],
+
+    // generators for $random* variables
+    dynamicGenerators = {
```
codenirvana: remove `dynamicGenerators` assignment and directly export the object.

test/unit/dynamic-variables.test.js
```diff
@@ -0,0 +1,111 @@
+var _ = require('lodash'),
+    dynamicVariables = require('../../lib/superstring/dynamic-variables'),
+    expect = require('chai').expect;
+
+describe('Dynamic variable', function () {
```
codenirvana: add test to check if the variable name (object key) starts with `$`

lib/superstring/dynamic-variables.js
coditva: We need to have the `dynamicGenerators` assignment because we are accessing `$randomDirectoryPath` inside `$randomFilePath`.

See: https://github.com/postmanlabs/postman-collection/pull/904/files/2029aa6b162b3662a3ad79291957cf89aa68a6b5#diff-b3d7c12136fa69f68a89843eb1b062a3R149

codenirvana: Got it, there's a workaround but it's fine. No change required.

PR #914, opened 2019-07-15
Add function `replaceIn` to replace variables in an object or string

coditva:
Ref: https://github.com/postmanlabs/postman-app-support/issues/2922

TODO:
- [x] Fix `[]` converting to `{}` and `['{{var}}']` to `{ 0: 'value' }`.

### DIFF
--- a/lib/collection/variable-scope.js
+++ b/lib/collection/variable-scope.js
@@ -316,7 +316,13 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
         this.values.clear();
     },
 
-    interpolate: function (obj) {
+    /**
+     * Replace all variable names with their values in the given string or object
+     *
+     * @param {string|object} obj A string or an object to replace variables in
+     * @returns {string|object} The string or object with variables (if any) substituted with their values
+     */
+    replaceIn: function (obj) {
         var transformed,
             result;
 
--- a/lib/collection/variable-scope.js
+++ b/lib/collection/variable-scope.js
@@ -316,6 +316,20 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
         this.values.clear();
     },
 
+    interpolate: function (obj) {
+        var transformed,
+            result;
+
+        if (typeof obj !== 'object') {
+            obj = [obj];
+            transformed = true;
+        }
+
+        result = Property.replaceSubstitutionsIn(obj, _.concat(this.values, this._layers));
+
+        return transformed ? result[0] : result;
+    },
+
     /**
      * Enable mutation tracking.
      *
--- a/lib/collection/variable-scope.js
+++ b/lib/collection/variable-scope.js
@@ -323,20 +323,17 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
      * @returns {string|object} The string or object with variables (if any) substituted with their values
      */
     replaceIn: function (template) {
-        var isString = typeof template === 'string',
-            result;
-
-        if (isString) {
-            template = [template];
+        if (_.isString(template) || _.isArray(template)) {
+            // convert template to object because replaceSubstitutionsIn only accepts objects
+            var result = Property.replaceSubstitutionsIn({ template: template }, _.concat(this.values, this._layers));
+            return result.template;
         }
 
-        if (typeof template !== 'object') {
-            return template;
+        if (_.isObject(template)) {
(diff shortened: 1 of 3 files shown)


coditva:
Rebased on `develop`.

lib/collection/variable-scope.js
```diff
@@ -316,6 +316,26 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
+     * Replace all variable names with their values in the given string or object
+     *
+     * @param {string|object} obj A string or an object to replace variables in
+     * @returns {string|object} The string or object with variables (if any) substituted with their values
+     */
+    replaceIn: function (obj) {
+        var transformed,
+            result;
+
+        if (typeof obj !== 'object') {
```
codenirvana: Instead of this, try:
- if `obj` is string, transform to an array
- if `obj` is not of type `object`, bail out
- finally, replace substitutions

This helps to avoid unnecessarily ops for undefined, null, etc. values.

lib/collection/variable-scope.js
```diff
@@ -316,6 +316,26 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
+            obj = [obj];
+            transformed = true;
+        }
+
+        result = Property.replaceSubstitutionsIn(obj, _.concat(this.values, this._layers));
```
codenirvana: test with multiple `layers`.

lib/collection/variable-scope.js
```diff
@@ -316,6 +316,29 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
...
+
+        if (isString) {
+            template = [template];
+        }
+
+        if (typeof template !== 'object') {
+            return template;
+        }
+
+        result = Property.replaceSubstitutionsIn(template, _.concat(this.values, this._layers));
```
shamasis: redo with this being called twice based on string case

lib/collection/variable-scope.js
```diff
@@ -316,6 +316,26 @@ _.assign(VariableScope.prototype, /** @lends VariableScope.prototype */ {
         this.values.clear();
     },
 
+    /**
+     * Replace all variable names with their values in the given template.
+     *
+     * @param {string|object} template - A string or an object to replace variables in
```
codenirvana:
```suggestion
     * @param {String|Object} template - A string or an object to replace variables in
```

test/unit/variable-scope.test.js
```diff
@@ -631,6 +631,194 @@ describe('VariableScope', function () {
...
+                        key: 'var1',
+                        value: 'value1'
+                    }]
+                });
+                expect(scope.replaceIn('{{var1}}')).to.equal('value1');
+                expect(scope.replaceIn(null)).to.equal(null);
+                expect(scope.replaceIn(undefined)).to.equal(undefined);
+                expect(scope.replaceIn(true)).to.equal(true);
+                expect(scope.replaceIn({})).to.eql({});
+                expect(scope.replaceIn('`console.error()`')).to.equal('`console.error()`');
```
codenirvana: Also, add Array and Function type

test/unit/variable-scope.test.js
```diff
@@ -631,6 +631,194 @@ describe('VariableScope', function () {
...
+                    values: [{
+                        key: 'name',
+                        value: 'Cooper'
+                    }, {
+                        key: 'job',
+                        value: 'Postman'
+                    }]
+                });
+
+                expect(scope.replaceIn('I am {{name}} and I work at {{job}}'))
```
codenirvana: Add test with nested variables (1 or 2 levels) as well.

test/unit/variable-scope.test.js
```diff
@@ -631,6 +631,194 @@ describe('VariableScope', function () {
...
+                        key: 'place',
+                        value: 'Bangalore'
+                    }]
+                });
+
+                expect(scope.replaceIn('{{place}} is not {{ place }}'))
+                    .to.equal('Bangalore is not {{ place }}');
+            });
+
+            it('should be able to resolve variables deeply', function () {
```
codenirvana: We don't use programmatic sets like this, better test such conditions individually.
We can remove this test.

test/unit/variable-scope.test.js
coditva: They are in separate tests. Do you want them in this test itself?

codenirvana: Yeah, not required.

PR #931, opened 2019-08-20
Fixed a bug where header with value `0` was being replaced by empty string

vikiCoder:
**Issue description:**
- Creating a new Header object with value `0` (as an Integer) was causing the value to be `''` (empty String) rather than having its original value.

**Fix:**
- Only set empty string as default value of the header if the provided value is `undefined` or `null`. Otherwise use provided value.

Issue: https://github.com/postmanlabs/postman-app-support/issues/4162


lib/collection/header.js
```diff
@@ -105,13 +105,13 @@ _.assign(Header.prototype, /** @lends Header.prototype */ {
          * @type {String}
          * @todo avoid headers with falsy key.
          */
-        this.key = _.get(options, 'key') || E;
+        this.key = _.isNil(_.get(options, 'key')) ? E : options.key;
```
codenirvana: Same as:

```suggestion
        this.key = _.get(options, 'key', E);
```

vikiCoder: No, it's different.
`_.get()` will not return default value for `null`.

PR #938, opened 2019-09-06
Handle malformed cookie value

codenirvana:
Handle [URIError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Malformed_URI) on decoding malformed cookie value.

Refer: https://github.com/postmanlabs/postman-app-support/issues/7268

**Discovered via** https://github.com/postmanlabs/postman-runtime/commit/e7c40eeaa37649a1cd8000b23548ed96435597e7.
Previously we used to emit POJO Cookie object instead of PostmanCookie instance in `response` callback of runtime.
Since https://github.com/postmanlabs/postman-runtime/pull/865, we construct a PostmanCookie List.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -58,8 +58,15 @@ _.mixin(/** @lends util */ {
      * @returns {string}
      */
     ensureEncoded: function (string) {
-        // Takes care of the case where the string is already encoded.
-        return encodeURIComponent(decodeURIComponent(string));
+        try {
+            // Takes care of the case where the string is already encoded.
+            return encodeURIComponent(decodeURIComponent(string));
+        }
+        // handle malformed URI sequence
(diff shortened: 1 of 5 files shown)


lib/util.js
```diff
@@ -58,8 +58,15 @@ _.mixin(/** @lends util */ {
...
-        // Takes care of the case where the string is already encoded.
-        return encodeURIComponent(decodeURIComponent(string));
+        try {
+            // Takes care of the case where the string is already encoded.
+            return encodeURIComponent(decodeURIComponent(string));
+        }
+        // handle malformed URI sequence
+        // refer: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Malformed_URI
+        catch (error) {
+            return string;
```
saswatds: Should this not be `encodeURIComponent(string)` ?

PR #960, opened 2019-11-28
Accept Response.stream to be a Base64 string

codenirvana:
Update `normalizeStream` to handle the response stream passed as base64 string.

### DIFF
--- a/lib/collection/response.js
+++ b/lib/collection/response.js
@@ -81,7 +81,21 @@ var util = require('../util'),
      * @const
      * @type {string}
      */
-    BUFFER = 'Buffer',
+    BASE64 = 'base64',
+
+    /**
+     * @private
+     * @const
+     * @type {string}
+     */
+    STREAM_TYPE_BUFFER = 'Buffer',
+
+    /**
+     * @private
+     * @const
+     * @type {string}
+     */
+    STREAM_TYPE_BASE64 = 'Base64',
 
     /**
      * @private
@@ -89,6 +103,13 @@ var util = require('../util'),
      * @type {string}
      */
     FUNCTION = 'function',
+
+    /**
+     * @private
+     * @const
+     * @type {string}
+     */
+    STRING = 'string',
 
     /**
      * @private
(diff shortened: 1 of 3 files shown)


lib/collection/response.js
```diff
@@ -142,17 +142,24 @@ var util = require('../util'),
...
-    normaliseStream = function (stream) {
+    normalizeStream = function (stream) {
+        // create buffer from buffer's JSON representation
         if (stream && stream.type === BUFFER && _.isArray(stream.data)) {
             // @todo Add tests for Browser environments, where ArrayBuffer is returned instead of Buffer
-            return typeof Buffer === FUNCTION ? new Buffer(stream.data) : new Uint8Array(stream.data).buffer;
+            return typeof Buffer === FUNCTION ? Buffer.from(stream.data) : new Uint8Array(stream.data).buffer;
         }
+        // create buffer from base64 string
+        else if (typeof stream === 'string') {
```
shamasis: Can you add comment here stating that based on the `if` statement condition structure, it is imperative that the string check lies in the ease block, otherwise blank string would be erroneously not converted to buffer. 

- [ ] add test for blank string stream check.

lib/collection/response.js
```diff
@@ -142,17 +142,24 @@ var util = require('../util'),
...
+            return Buffer.from(stream, 'base64');
```
shamasis: Default stream encoding should be put up as a constant.

shamasis: There should be an identification saying the stream is base64 somehow before we default decode. Reason being that otherwise the default would prevent us from modifying stream structure and yet retain compatibility with older SDK running in older apps.

lib/collection/response.js
```diff
@@ -189,8 +196,9 @@ _.inherit((
 
 _.assign(Response.prototype, /** @lends Response.prototype */ {
     update: function (options) {
-        // options.stream accepts new Buffer() as well as new Buffer().toJSON()
-        var stream = normaliseStream(options.stream);
+        // options.stream accepts Buffer, Buffer.toJSON() or base64 string
+        // @todo this doubles the memory footprint (options.stream + generated buffer).
```
shamasis: Do you mean it doubles the footprint or you mean that it causes temporary consumption spike?

lib/collection/response.js
shamasis: Worst case there should be a way to indicate to not process the stream in case we choose any other format in future.

shamasis: What got in the way of doing dataURI equivalent?

lib/collection/response.js
codenirvana: Yes, causes temporary consumption spike.

lib/collection/response.js
codenirvana: The data URI scheme (`data:[<media type>][;base64],<data>`) comes with the overhead of string manipulation for media type, whitespace inconsistency between `[;base64],` and `<data>` etc.

The best way to handle all that is by using RegEx which will hit performance for large response payload. And, cause a temporary memory consumption spike for the additional strings constructed.

I would rather prefer an object representation instead of embracing the limitations of the HTML attribute (as data URI is built for that).
```js
{
    encoding: "base64",
    data: "ZGF0YQ==",
    mimeInfo: {}
}
```

codenirvana: The current approach enforces stream property to be either a valid:
- Buffer
- Buffer.toJSON()
- Buffer.toString('base64')

PR #984, opened 2020-02-11
Retain empty url segments on toString()

codenirvana:
Handles cases like:
```js
it('should retain @ in auth without user and password', function () {
    var url = 'http://@localhost';
    expect((new Url(url)).toString()).to.equal(url);
});

it('should retain : in empty port', function () {
    var url = 'localhost:/path';
    expect((new Url(url)).toString()).to.equal(url);
});

it('should retain # in empty hash', function () {
    var url = 'localhost#';
    expect((new Url(url)).toString()).to.equal(url);
});

it('should retain ? in empty query param', function () {
    var url = 'http://localhost?';
    expect((new Url(url)).toString()).to.equal(url);
});
```

Also added tests for `Url~toString`.

### DIFF
--- a/test/unit/url.test.js
+++ b/test/unit/url.test.js
@@ -1082,6 +1082,11 @@ describe('Url', function () {
             expect((new Url(url)).toString()).to.equal(url);
         });
 
+        it('should retain : in auth with empty user and password', function () {
+            var url = 'http://:@localhost';
+            expect((new Url(url)).toString()).to.equal(url);
+        });
+
         it('should retain : in empty port', function () {
             var url = 'localhost:/path';
             expect((new Url(url)).toString()).to.equal(url);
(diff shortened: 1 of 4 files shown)


test/unit/url.test.js
```diff
@@ -1317,6 +1473,20 @@ describe('Url', function () {
...
+            it('should handle query parameters with empty key or value', function () {
+                var url = new Url('https://postman-echo.com?get&=bar&=&baz=&&');
+
+                expect(url.query.all()).to.have.deep.members([
+                    new QueryParam({ key: 'get', value: null }),
+                    new QueryParam({ key: '', value: 'bar' }),
+                    new QueryParam({ key: '', value: '' }),
+                    new QueryParam({ key: 'baz', value: '' }),
+                    new QueryParam({ key: null, value: null }),
+                    // why's ☝🏻 different from 👇🏻 (╯°□°)╯︵ ┻━┻
```
vikiCoder: It is because of this: https://github.com/postmanlabs/postman-collection/pull/984/files#diff-e02ef5f2975cda047ffc0d6996253ae0R120

But why is it needed!? 🤔

codenirvana: Yes, but don't want to change that flow right now.

PR #1012, opened 2020-03-25
URLencoding using postman-url-encoder

ProBeta1:


codenirvana:
Hey @ProBeta1, thanks for the contribution but this `@todo` is no longer relevant and hence the change is not required because the long term plan is to remove the `postman-url-encoder` dependency completely from this library.

ProBeta1:
Okay , was not aware of that !

PR #1013, opened 2020-03-27
proxy-config schema updated

someshkoli:
- Removed `server` property
 - Added the following properties
       - `username` 
       - `password`    
       - `host`
       - `port`
       - `bypass`
       - `authenticate`


codenirvana:
Schema is currently maintained here https://github.com/postmanlabs/schemas.

PR #1060, opened 2020-06-21
Add $isoTimestamp to dynamic variables

atulbi:
Many APIs need timestamps to be send with the request and using ISO format is common to use. Currently, the dynamic variable only has `$timestamp` variable which provides the number of milliseconds since epoch.
Adding `$isoTimestamp` will provide the timestamp in ISO format at zero UTC.

Name: $isoTimestamp
Description: _The current ISO timestamp at zero UTC_

### DIFF
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -1,3 +1,7 @@
+master:
+  new features:
+    - GH-1060 Added $isoTimestamp to the dynamic variable list
+
 3.6.2:
   date: 2020-06-04
   fixed bugs:
--- a/lib/superstring/dynamic-variables.js
+++ b/lib/superstring/dynamic-variables.js
@@ -91,6 +91,13 @@ var faker = require('faker/locale/en'),
             }
         },
 
+        $isoTimestamp: {
+            description: 'The current ISO timestamp at zero UTC',
+            generator: function () {
+                return new Date().toISOString();
+            }
+        },
+
         $randomInt: {
             description: 'A random integer between 1 and 1000',
             generator: function () {
--- a/test/unit/dynamic-variables.test.js
+++ b/test/unit/dynamic-variables.test.js
@@ -33,6 +33,17 @@ describe('Dynamic variable', function () {
             expect(dynamicVariables.$timestamp.generator()).to.be.a('number');
         });
 
+        it('$isoTimestamp returns a timestamp in ISO format', function () {
+            var isoTimestamp = dynamicVariables.$isoTimestamp.generator(),
+                // eslint-disable-next-line security/detect-non-literal-regexp
+                regex = new RegExp('^(-?(?:[1-9][0-9]*)?[0-9]{4})' + // match year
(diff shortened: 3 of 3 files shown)


test/unit/dynamic-variables.test.js
```diff
@@ -86,5 +86,14 @@ describe('Dynamic variable', function () {
...
             expect(directoryPath).to.not.be.null;
         });
+
+        it('$isoTimestamp returns a timestamp in ISO format', function () {
+            var isoTimestamp = dynamicVariables.$isoTimestamp.generator();
+
+            expect(isoTimestamp).to.be.a('string');
+            expect(isoTimestamp).to
+                // eslint-disable-next-line max-len
+                .match(/^([+-]?\d{4}(?!\d{2}\b))((-?)((0[1-9]|1[0-2])(\3([12]\d|0[1-9]|3[01]))?|W([0-4]\d|5[0-2])(-?[1-7])?|(00[1-9]|0[1-9]\d|[12]\d{2}|3([0-5]\d|6[1-6])))([T\s]((([01]\d|2[0-3])((:?)[0-5]\d)?|24:?00)([.,]\d+(?!:))?)?(\17[0-5]\d([.,]\d+)?)?([zZ]|([+-])([01]\d|2[0-3]):?([0-5]\d)?)?)?)?$/);
```
coditva, line 96: Is there a way to simplify this regex? This is unreadable and we have no idea what/how it's testing.

lib/superstring/dynamic-variables.js
```diff
@@ -602,6 +602,12 @@ var faker = require('faker/locale/en'),
         $randomSemver: {
             description: 'A random semantic version number',
             generator: faker.system.semver
+        },
+        $isoTimestamp: {
+            description: 'The current ISO timestamp at zero UTC',
```
coditva, line 607: Have we gotten the variable name and copy approved?

CHANGELOG.yaml
```diff
@@ -1,3 +1,7 @@
+master:
+  new features:
+    - Added $isoTimestamp to the dynamic variable list
```
coditva, line 3: Let's add the GitHub issue ID here as well: https://github.com/postmanlabs/postman-app-support/issues/553

lib/superstring/dynamic-variables.js
```diff
@@ -602,6 +602,12 @@ var faker = require('faker/locale/en'),
         $randomSemver: {
             description: 'A random semantic version number',
             generator: faker.system.semver
+        },
+        $isoTimestamp: {
```
codenirvana, line 606: Move it closer to `$timestamp`.

test/unit/dynamic-variables.test.js
atulbi: This covers many time-related corner cases. 
Let's update it with a simpler one. I will also add comments to it.

IanKemp:
When can we expect an updated version of the Postman app with this change? It has been nearly 2 weeks now.

atulbi:
App will be updated in 7.29 with these changes.

PR #1081, opened 2020-08-19
Checking whether key is callable or not

sadakchap:
It is fix for this [issue](https://github.com/postmanlabs/postman-app-support/issues/8924) .

The issue was caused when we had query parameter key named "hasOwnProperty" which is already a defined key in any object whose value is a function that can be used to check whether a key exists or not.

When adding any query parameter, we also check whether multiple values are allowed for that query parameter key, if allowed than we look for that key using "hasOwnProperty" function and if found than push query parameter value, else we create query parameter key with value. 
But when we have "hasOwnProperty" as query parameter key and try to add that to PostmanPropertyList object, it overwrites that default function value, and then we can not further call "hasOwnProperty" to check for a key in an object.

Here, I have added a check to see if the key is callable, if it is callable, then return function call, else if key's value is not undefined then push value to existing key, else add new query parameter key.
image

### DIFF
--- a/lib/collection/property-list.js
+++ b/lib/collection/property-list.js
@@ -132,7 +132,7 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
             // if multiple values are allowed, the reference may contain an array of items, mapped to an index.
             if (this._postman_listAllowsMultipleValues && Object.hasOwnProperty.call(this.reference, index)) {
 
-                // if the value is not an array, convert it to an array.
+                // if the value is not an array, convert it to an array
                 !_.isArray(this.reference[index]) && (this.reference[index] = [this.reference[index]]);
 
                 // add the item to the array of items corresponding to this index
--- a/lib/collection/property-list.js
+++ b/lib/collection/property-list.js
@@ -130,7 +130,9 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
             this._postman_listIndexCaseInsensitive && (index = index.toLowerCase());
 
             // if multiple values are allowed, the reference may contain an array of items, mapped to an index.
-            if (this._postman_listAllowsMultipleValues && this.reference.hasOwnProperty(index)) {
+            if (this._postman_listAllowsMultipleValues &&
+                _.isFunction(this.reference.hasOwnProperty) ?
+                this.reference.hasOwnProperty(index) : this.reference.index !== undefined) {
 
                 // if the value is not an array, convert it to an array.
                 !_.isArray(this.reference[index]) && (this.reference[index] = [this.reference[index]]);
--- a/lib/collection/property-list.js
+++ b/lib/collection/property-list.js
@@ -132,7 +132,7 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
             // if multiple values are allowed, the reference may contain an array of items, mapped to an index.
             if (this._postman_listAllowsMultipleValues && Object.hasOwnProperty.call(this.reference, index)) {
 
-                // if the value is not an array, convert it to an array
+                // if the value is not an array, convert it to an array.
                 !_.isArray(this.reference[index]) && (this.reference[index] = [this.reference[index]]);
 
                 // add the item to the array of items corresponding to this index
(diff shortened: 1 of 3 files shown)


lib/collection/property-list.js
```diff
@@ -130,7 +130,9 @@ _.assign(PropertyList.prototype, /** @lends PropertyList.prototype */ {
             this._postman_listIndexCaseInsensitive && (index = index.toLowerCase());
 
             // if multiple values are allowed, the reference may contain an array of items, mapped to an index.
-            if (this._postman_listAllowsMultipleValues && this.reference.hasOwnProperty(index)) {
+            if (this._postman_listAllowsMultipleValues &&
```
codenirvana, line 133: If you check MDN docs for [hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#Using_hasOwnProperty_as_a_property_name), it has an example for handling this situation where you call `Object.prototype.hasOwnProperty` directly.

codenirvana (changes requested):
Can we also add a unit test for this change?

lib/collection/property-list.js
sadakchap: okay. I have tried adding a unit test [here](https://github.com/sadakchap/postman-collection/blob/20a38ac9c0c695162f6246ee1441490d91941f90/test/unit/url.test.js#L1330).

 And will call `Object.prototype.hasOwnProperty ` directly instead of checking whether key is callable or not. 😀

test/unit/url.test.js
```diff
@@ -1327,6 +1327,16 @@ describe('Url', function () {
             ]);
         });
 
+        it('should correctly add hasOwnProperty query key to an existing Url instance', function () {
```
codenirvana, line 1330: For context, let's add link to the reported issue.
```suggestion
        // Refer: https://github.com/postmanlabs/postman-app-support/issues/8924
        it('should correctly add hasOwnProperty query key to an existing Url instance', function () {
```

codenirvana (changes requested):
Looks good to me 👍 
But, can we add an isolated test as well for the `PropertyList.insert` method?

lib/collection/property-list.js
codenirvana: 👍  Testing the query parameter covers the reported bug but let's also add an isolated (unit) test which tests this change only i.e, a unit test for the `PropertyList.insert` method.

test/unit/url.test.js
sadakchap: About adding an isolate test for `PropertyList.insert` method. I tried adding a test there and could not get that test to fail. It was passing fine.
The reason is, In the `property-list.test.js` file, when making an instance of PropertyList we do not pass any parameters to the constructor, due to this `_postman_listAllowsMultipleValues` is set to false and due to short circuit [this condition](https://github.com/postmanlabs/postman-collection/blob/043e8b1bb95d48ad2854485b051a47ec90d7d02c/lib/collection/property-list.js#L133), it never checks for remaining && condition (`this.reference.hasOwnProperty(index)`) and also never checks members with multiple values key. That's why test will never fail there.

Sure, can change test display message.

codenirvana: You can create a `PropertyList` with a custom type having `_postman_propertyAllowsMultipleValues` set to true.
```js
var FakeType,
    list;

FakeType = function (options) {
    this.key = options.key;
    this.value = options.value;
};

FakeType._postman_propertyIndexKey = 'key';
FakeType._postman_propertyAllowsMultipleValues = true;

list = new sdk.PropertyList(FakeType, {}, { key: 'hasOwnProperty', value: 'false' });
list.insert({ key: 'foo', value: 'bar' }); // Uncaught TypeError: this.reference.hasOwnProperty is not a function
```

sadakchap: ok, As you have suggested I have tried adding test for `PropertyList.insert()` method [here](https://github.com/postmanlabs/postman-collection/pull/1081/commits/0b7d4078c1a59bd7aa4ebb54a11919f1d72b2c2f#diff-c6d14c63e95734583772196465ecded5) .

PR #1098, opened 2020-09-03
Adding unit tests

lucifercr07:
Unit tests being added for the following files:
- cookie.js
- header.js
- request-auth.js
- item.js
- content-info.js

### DIFF
--- a/test/unit/content-info.test.js
+++ b/test/unit/content-info.test.js
@@ -174,6 +174,28 @@ describe('contentInfo module', function () {
             });
     });
 
+    it('Should return with default file name if invalid charset type specified', function () {
+        var response = new Response({ header: [
+            {
+                key: 'Content-Disposition',
+                value: 'attachment;filename*=invalid-charset\'\'%E5%93%8D%E5%BA%94.json'
+            },
+            {
+                key: 'Content-Type',
+                value: 'application/json'
+            }
+        ], stream: Buffer.from('a test json').toJSON()
+        });
+        expect(contentInfo.contentInfo(response)).to
+            .eql({
+                charset: 'utf8',
+                fileExtension: 'json',
+                mimeFormat: 'json',
+                mimeType: 'text',
+                fileName: 'response.json'
+            });
+    });
+
     it('Should return file name from content-disposition header with encoded type utf-8', function () {
         var response = new Response({ header: [
             {
--- a/test/unit/content-info.test.js
+++ b/test/unit/content-info.test.js
@@ -24,16 +24,16 @@ describe('contentInfo module', function () {
     });
 
     it('Should return file name from content-disposition header with type attachment and file name', function () {
-        var response = new Response({ header: [
-            {
+        var response = new Response({
+            header: [{
                 key: 'content-disposition',
                 value: 'attachment; filename=testResponse.json'
             },
             {
                 key: 'content-type',
                 value: 'application/json'
-            }
-        ], stream: Buffer.from('random').toJSON()
+            }],
+            stream: Buffer.from('random').toJSON()
         });
         expect(contentInfo.contentInfo(response)).to.eql({
             charset: 'utf8',
@@ -45,12 +45,12 @@ describe('contentInfo module', function () {
     });
 
     it('Should return default file name and extension from mime if disposition header is not present ', function () {
-        var response = new Response({ header: [
-            {
+        var response = new Response({
+            header: [{
                 key: 'content-type',
                 value: 'application/json'
-            }
-        ], stream: Buffer.from('random').toJSON()
+            }],
+            stream: Buffer.from('random').toJSON()
         });
         expect(contentInfo.contentInfo(response)).to.eql({
             charset: 'utf8',
@@ -63,16 +63,16 @@ describe('contentInfo module', function () {
 
     // eslint-disable-next-line max-len
     it('Should return filename from content-disposition header with type attachment and file name without ext', function () {
-        var response = new Response({ header: [
-            {
+        var response = new Response({
+            header: [{
                 key: 'Content-Disposition',
                 value: 'attachment; filename=testResponse'
             },
             {
                 key: 'Content-Type',
                 value: 'application/json'
-            }
-        ], stream: Buffer.from('a test json').toJSON()
+            }],
+            stream: Buffer.from('a test json').toJSON()
         });
         expect(contentInfo.contentInfo(response)).to
             .eql({
@@ -86,16 +86,16 @@ describe('contentInfo module', function () {
 
     // eslint-disable-next-line max-len
     it('Should return file name from content-disposition header with type attachment and file name with dots', function () {
-        var response = new Response({ header: [
-            {
+        var response = new Response({
+            header: [{
                 key: 'Content-Disposition',
                 value: 'attachment;filename=test.Response.json'
             },
             {
                 key: 'Content-Type',
                 value: 'application/json'
-            }
-        ], stream: Buffer.from('a test json').toJSON()
+            }],
+            stream: Buffer.from('a test json').toJSON()
         });
         expect(contentInfo.contentInfo(response)).to
             .eql({
@@ -108,16 +108,16 @@ describe('contentInfo module', function () {
     });
 
     it('Should return file name from content-disposition header with type attachment and dot files', function () {
-        var response = new Response({ header: [
-            {
+        var response = new Response({
+            header: [{
                 key: 'Content-Disposition',
                 value: 'attachment;filename=.response'
             },
             {
                 key: 'Content-Type',
                 value: 'application/json'
-            }
-        ], stream: Buffer.from('a test json').toJSON()
+            }],
+            stream: Buffer.from('a test json').toJSON()
         });
         expect(contentInfo.contentInfo(response)).to
             .eql({
@@ -130,17 +130,17 @@ describe('contentInfo module', function () {
     });
 
     // eslint-disable-next-line max-len
-    it('Should return file name from content-disposition header with type attachment and file name with quotes', function () {
-        var response = new Response({ header: [
-            {
+    it('should return file name from content-disposition header with type attachment and file name with quotes', function () {
+        var response = new Response({
+            header: [{
                 key: 'Content-Disposition',
                 value: 'attachment;filename="test Response.json"'
             },
             {
                 key: 'Content-Type',
                 value: 'application/json'
-            }
-        ], stream: Buffer.from('a test json').toJSON()
+            }],
+            stream: Buffer.from('a test json').toJSON()
         });
         expect(contentInfo.contentInfo(response)).to
             .eql({
@@ -152,17 +152,17 @@ describe('contentInfo module', function () {
             });
     });
 
-    it('Should return file name from content-disposition header with type inline and file name', function () {
-        var response = new Response({ header: [
-            {
+    it('should return file name from content-disposition header with type inline and file name', function () {
+        var response = new Response({
+            header: [{
                 key: 'Content-Disposition',
                 value: 'attachment;filename=test.Response.json'
             },
             {
                 key: 'Content-Type',
                 value: 'application/json'
-            }
-        ], stream: Buffer.from('a test json').toJSON()
+            }],
+            stream: Buffer.from('a test json').toJSON()
         });
         expect(contentInfo.contentInfo(response)).to
             .eql({
@@ -174,17 +174,17 @@ describe('contentInfo module', function () {
             });
     });
 
-    it('Should return default file name if charset specified in content-disposition header is invalid', function () {
-        var response = new Response({ header: [
-            {
+    it('should return default file name if charset specified in content-disposition header is invalid', function () {
+        var response = new Response({
+            header: [{
                 key: 'Content-Disposition',
                 value: 'attachment;filename*=invalid-charset\'\'%E5%93%8D%E5%BA%94.json'
             },
             {
                 key: 'Content-Type',
                 value: 'application/json'
-            }
-        ], stream: Buffer.from('a test json').toJSON()
+            }],
+            stream: Buffer.from('a test json').toJSON()
         });
         expect(contentInfo.contentInfo(response)).to
             .eql({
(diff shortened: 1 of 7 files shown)


test/unit/cookie.test.js
```diff
@@ -194,4 +194,135 @@ describe('Cookie', function () {
...
+        });
+    });
+
+    describe('unparseSingle', function () {
+        it('should return only the value in case of no cookie name specified', function () {
+            var rawCookie = {
+                    expires: 1502442248,
+                    hostOnly: false,
+                    httpOnly: false,
+                    key: 'foo',
```
coditva, line 213: We have specified a `key` here. Shouldn't it be used for `name`? 🤔

coditva, line 213: I think it's worth adding a test to check that when `name` is not specified, `key` is used instead.

test/unit/cookie.test.js
```diff
@@ -194,4 +194,135 @@ describe('Cookie', function () {
...
+                unparse;
+            cookieArray.push(rawCookie1);
+            cookieArray.push(rawCookie2);
+
+            unparse = Cookie.unparse(cookieArray);
+            expect(unparse).equals('testCookie1=fooTest1; testCookie2=fooTest2');
+        });
+    });
+
+    describe('stringify', function () {
```
coditva, line 268: Shouldn't we add tests for `stringify` similar to all tests for `toString`?

test/unit/cookie.test.js
```diff
@@ -194,4 +194,135 @@ describe('Cookie', function () {
...
+            var cookie = new Cookie({
+                key: 'foo',
+                value: 'fooTest',
+                maxAge: 1502442248
+            });
+            expect(cookie.toString())
+                .equals('foo=fooTest; Max-Age=1502442248');
+        });
+    });
 });
```
coditva, line 328: Should we also add tests for cases when each of the options are invalid? `null`, `object` instead of `string`, etc.

test/unit/cookie.test.js
```diff
@@ -194,4 +194,135 @@ describe('Cookie', function () {
...
+                    hostOnly: false,
+                    httpOnly: false,
+                    key: 'foo',
+                    path: '/',
+                    secure: false,
+                    session: false,
+                    value: 'bar'
+                },
+                unparsedSingle = Cookie.unparseSingle(rawCookie);
+            expect(unparsedSingle).equals('bar');
```
coditva, line 220: Let's change `.equals` to `.to.equal` because it is more readable and forms a sentence:
> expect **actual value** to equal **expected value**

test/unit/header.test.js
```diff
@@ -70,6 +70,15 @@ describe('Header', function () {
                 value: 0
             });
         });
+
+        it('should create the header with system options', function () {
+            var header = Header.create({ key: 'name', value: 'value', system: true });
+            expect(header.toJSON()).to.eql({
+                key: 'name',
+                value: 'value',
+                system: true
```
coditva, line 79: What happens when `system` is not boolean?

test/unit/header.test.js
```diff
@@ -70,6 +70,15 @@ describe('Header', function () {
                 value: 0
             });
         });
```
coditva, line 72: Let's add tests for invalid values of `key` as well.

test/unit/header.test.js
```diff
@@ -229,4 +238,13 @@ describe('Header', function () {
...
     });
+
+    describe('toString', function () {
+        it('should convert the header to a single header string', function () {
+            // eslint-disable-next-line max-len
+            var rawHeader = 'Content-Type: application/json\nAuthorization: Hawk id="dh37fgj492je", ts="1448549987", nonce="eOJZCd", mac="O2TFlvAlMvKVSKOzc6XkfU6+5285k5p3m5dAjxumo2k="',
+                newHeader = new Header(rawHeader);
+            expect(newHeader.toString()).equals(rawHeader);
+        });
+    });
```
coditva, line 249: Let's also add a test where we create a new header by passing an object to `new Header()` and check the `toString` output.

test/unit/item.test.js
```diff
@@ -466,6 +466,27 @@ describe('Item', function () {
...
             expect(item.toJSON()).to.have.property('protocolProfileBehavior').that.is.empty;
         });
+
+        it('should not unset protocolProfileBehavior from an Item incase key is not of type string', function () {
+            var item = new Item({
+                protocolProfileBehavior: { keyName: null }
+            });
+
+            item.unsetProtocolProfileBehavior(null);
+            expect(item.toJSON()).to.have.property('protocolProfileBehavior').that.is.not.empty;
```
coditva, line 476: Let's also assert on the value of `item.protocolProfileBehavior`

test/unit/item.test.js
```diff
@@ -466,6 +466,27 @@ describe('Item', function () {
             item.unsetProtocolProfileBehavior('keyName');
             expect(item.toJSON()).to.have.property('protocolProfileBehavior').that.is.empty;
         });
+
+        it('should not unset protocolProfileBehavior from an Item incase key is not of type string', function () {
+            var item = new Item({
+                protocolProfileBehavior: { keyName: null }
```
coditva, line 472: Why is the value `null`? It makes the test a little confusing.

test/unit/request-auth.test.js
```diff
@@ -4,123 +4,185 @@ var expect = require('chai').expect,
...
+                    foo: 'bar'
+                },
+                type: 'noauth'
+            });
 
-        auth.use('basic');
-        expect(auth.type).to.equal('basic');
-        expect(auth.current()).to.eql({
-            username: 'u', password: 'p'
+            auth.use('noauth', {
```
coditva, line 87: Why are we setting the auth options again in the `.use` method? The test name does not say anything about it.
It is making the `expect` a little confusing.

If we're testing that `.use` can update the options, I think we can break this test into two.

test/unit/request-auth.test.js
```diff
@@ -4,123 +4,185 @@ var expect = require('chai').expect,
...
+                noauth: {
+                    foo: 'bar'
+                },
+                type: 'noauth'
+            });
 
-        expect(auth.current()).to.eql({});
+            auth.update(null);
+            auth.update({});
+            expect(auth.parameters().toObject()).to.eql({
```
coditva, line 130: It would make more sense to either break this test into multiple tests for each invalid value, or add expect after each `.update`.

We should also check for other invalid values like `string`, `numbers`, `boolean`, etc.

coditva (changes requested):

coditva (changes requested):
- [x] Also need to update `CHANGELOG.yaml`

test/unit/request-auth.test.js
lucifercr07: Fixed

test/unit/request-auth.test.js
lucifercr07: It's the latter one, have updated the test description to depict that we are updating the auth via options.

test/unit/cookie.test.js
coditva: I think this might be a bug? We allow using `key` in place of `name` when creating a cookie. Ref: [/lib/collection/cookie.js#L125](https://github.com/postmanlabs/postman-collection/blob/16cc6f821a2af99d979477f9d97432b1d5362090/lib/collection/cookie.js#L125).

@codenirvana views?

test/unit/request-auth.test.js
coditva: I think it's better for readability that we pass just `{ foo1: 'bar1' }` rather than the whole thing again.
It also makes the usage of the function more clear.

coditva (changes requested):

test/unit/cookie.test.js
coditva: 🎉

test/unit/cookie.test.js
```diff
@@ -287,19 +288,63 @@ describe('Cookie', function () {
...
 
     describe('stringify', function () {
+        it('should return single Set-Cookie header string with key value', function () {
+            var cookie = new Cookie({
+                name: 'foo',
+                value: 'fooTest'
+            });
+
+            expect(Cookie.stringify(cookie))
+                .equals('foo=fooTest');
```
coditva, line 298: Nitpick: `.to.equals` allows better readability. And we can keep it on the same line.

It's your choice to change it in this PR. But we can keep in mind for later PRs.

coditva (dismissed):
LGTM! 🚀

test/unit/cookie.test.js
```diff
@@ -46,6 +46,34 @@ describe('Cookie', function () {
...
+                    hostOnly: false,
+                    httpOnly: false,
+                    path: '/',
+                    secure: false,
+                    session: false,
+                    value: 'fooTest'
+                }),
+                unparsedSingle = Cookie.unparseSingle(testCookie);
+
+            expect(unparsedSingle).to.equal(testCookie.name + '=' + testCookie.value);
```
codenirvana, line 74: Assert the exact string value instead of concatenating.

test/unit/cookie.test.js
```diff
@@ -193,5 +221,178 @@ describe('Cookie', function () {
...
+                    expires: 1502442248,
+                    hostOnly: false,
+                    httpOnly: false,
+                    path: '/',
+                    secure: false,
+                    session: false,
+                    value: 'fooTest'
+                },
+                unparsedSingle = Cookie.unparseSingle(new Cookie(rawCookie));
+            expect(unparsedSingle).to.equal(rawCookie.name + '=' + rawCookie.value);
```
codenirvana, line 246: Assert the exact string.

test/unit/cookie.test.js
```diff
@@ -193,5 +221,178 @@ describe('Cookie', function () {
...
+                    value: 'fooTest1'
+                },
+                rawCookie2 = {
+                    name: 'testCookie2',
+                    domain: '.httpbin.org',
+                    expires: 1502442248,
+                    path: '/',
+                    value: 'fooTest2'
+                },
+                cookieArray = [],
```
codenirvana, line 280:
```suggestion
                cookieArray = [rawCookie1, rawCookie2],
```

test/unit/cookie.test.js
```diff
@@ -193,5 +221,178 @@ describe('Cookie', function () {
...
+
+    describe('stringify', function () {
+        it('should return single Set-Cookie header string with key value', function () {
+            var cookie = new Cookie({
+                name: 'foo',
+                value: 'fooTest'
+            });
+
+            expect(Cookie.stringify(cookie))
+                .equals('foo=fooTest');
```
codenirvana, line 298:
```suggestion
            expect(Cookie.stringify(cookie)).to.equals('foo=fooTest');
```

test/unit/cookie.test.js
```diff
@@ -193,5 +221,178 @@ describe('Cookie', function () {
...
+        });
+
+        it('should return single Set-Cookie header string with expires', function () {
+            var cookie = new Cookie({
+                name: 'foo',
+                value: 'fooTest',
+                expires: '14 Jun 2017 00:00:00 PDT'
+            });
+
+            expect(Cookie.stringify(cookie))
```
codenirvana, line 308: `.to. equals`

test/unit/cookie.test.js
```diff
@@ -193,5 +221,178 @@ describe('Cookie', function () {
...
+        it('should return single Set-Cookie header string with secure and httpOnly set to true', function () {
+            var cookie = new Cookie({
+                name: 'foo',
+                value: 'fooTest',
+                httpOnly: true,
+                secure: true
+            });
+
+            expect(Cookie.stringify(cookie))
+                .equals('foo=fooTest; Secure; HttpOnly');
```
codenirvana, line 321: Split into multiple lines only if it exceeds 120 chars limit.

test/unit/content-info.test.js
```diff
@@ -174,6 +174,28 @@ describe('contentInfo module', function () {
             });
     });
 
+    it('Should return default file name if charset specified in content-disposition header is invalid', function () {
```
codenirvana, line 177:
```suggestion
    it('should return default file name if charset specified in content-disposition header is invalid', function () {
```

test/unit/content-info.test.js
```diff
@@ -174,6 +174,28 @@ describe('contentInfo module', function () {
...
+            {
+                key: 'Content-Disposition',
+                value: 'attachment;filename*=invalid-charset\'\'%E5%93%8D%E5%BA%94.json'
+            },
+            {
+                key: 'Content-Type',
+                value: 'application/json'
+            }
+        ], stream: Buffer.from('a test json').toJSON()
+        });
```
codenirvana, line 188:
```suggestion
        var response = new Response({
            header: [{
                key: 'Content-Disposition',
                value: 'attachment;filename*=invalid-charset\'\'%E5%93%8D%E5%BA%94.json'
            }, {
                key: 'Content-Type',
                value: 'application/json'
            }],
            stream: Buffer.from('a test json').toJSON()
        });
```

fix formatting.

codenirvana (changes requested):

test/unit/content-info.test.js
lucifercr07: Fixed.

test/unit/content-info.test.js
lucifercr07: Fixed.

test/unit/cookie.test.js
lucifercr07: Fixed.

test/unit/cookie.test.js
lucifercr07: Fixed.

test/unit/cookie.test.js
lucifercr07: Fixed.

test/unit/cookie.test.js
lucifercr07: Fixed.

test/unit/cookie.test.js
lucifercr07: Fixed.

test/unit/cookie.test.js
lucifercr07: Fixed.

test/unit/cookie.test.js
lucifercr07: Fixed.

test/unit/cookie.test.js
codenirvana: Yes, this looks like a mistake but it's fine for now. I will take a look at this.

codenirvana (approved):

PR #1101, opened 2020-09-04
Fix for issues in request-auth file

lucifercr07:
- request-auth.js:
Issue - In case of an invalid auth type update was setting auth to default from the existing type.
Fix: Valid behavior should be that in case of invalid auth type update() should ideally return.

Added unit-tests corresponding to the changes.


coditva:
@lucifercr07 Update the `CHANGELOG.yaml` as well. In this case, an entry should be added under chores as:

```yaml
master:
  chores:
    - GH-1101 A small description about change
```

CHANGELOG.yaml
```diff
@@ -1,3 +1,10 @@
+3.6.6:
```
coditva: This should be `master` because the change is not released yet. Once released, this is changed to the version by @codenirvana 

```suggestion
master:
```

CHANGELOG.yaml
```diff
@@ -1,3 +1,10 @@
...
+  date: 2020-09-04
```
coditva, line 2: No needed for `master`.

```suggestion
```

CHANGELOG.yaml
```diff
@@ -1,3 +1,10 @@
...
+  fixed bugs:
+    - >-
+      Fixed a issue where in case of an invalid auth type update() was setting 
+      auth to default from the existing type.
```
coditva, line 6: Seems a little confusing, how about something along the lines of this:

```suggestion
      Fixed an issue in RequestAuth where type was set to previous auth type when update() called with invalid type.
```

coditva (changes requested):

coditva:
LGTM! 🚀

coditva (dismissed):

lib/collection/request-auth.js
```diff
@@ -74,8 +74,8 @@ _.assign(RequestAuth.prototype, /** @lends RequestAuth.prototype */ {
     update: function (options, type) {
         // update must have options
         if (!_.isObject(options)) { return; }
-        // validate type parameter and/or choose default from existing type.
-        if (!RequestAuth.isValidType(type)) { type = this.type; }
+        // validate type parameter and return in case type is not valid.
+        if (!RequestAuth.isValidType(type)) { return; }
```
codenirvana, line 78: It should default to the current `this.type` if not provided but bail out if an invalid type is provided.

If this change is not failing any test, let's make sure there's a test for it.

codenirvana (changes requested):

CHANGELOG.yaml
```diff
@@ -1,3 +1,9 @@
+master:
+  fixed bugs:
+    - >-
+      Fixed an issue in RequestAuth where type was set to previous auth type
```
codenirvana, line 4: As mentioned in https://github.com/postmanlabs/postman-collection/pull/1101#issuecomment-686995373, why's PR id missing here?

codenirvana (changes requested):

lib/collection/request-auth.js
lucifercr07: Fixed and added test.

CHANGELOG.yaml
lucifercr07: Fixed.

test/unit/request-auth.test.js
```diff
@@ -123,4 +123,85 @@ describe('RequestAuth', function () {
             fakeParent: true
         });
     });
+
+    describe('.update', function () {
```
codenirvana, line 127: If I have to follow TDD then I would like this test suite to be:

```
#update
- should update the parameters of a specific authentication type
- should update the parameters of the existing type if not specified explicitly
- should bail out if invalid auth type is provided
```

codenirvana (changes requested):

test/unit/request-auth.test.js
lucifercr07: Fixed.

codenirvana (dismissed):

codenirvana (approved):

PR #1104, opened 2020-09-08
Fix for cookie extension issue.

lucifercr07:
Issue: cookie extensions not getting stringified properly. Fixed check for `path` and `domain`.

### DIFF
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -1,5 +1,6 @@
 master:
   chores:
+    - GH-1104 Fixed cookie extensions not getting stringified properly. Fixed check for path and domain.
     - GH-1098 Added unit tests for cookie.js, header.js, request-auth.js, item.js, content-info.js.
 
 3.6.6:
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -1,6 +1,8 @@
 master:
   chores:
-    - GH-1104 Fixed cookie extensions not getting stringified properly. Fixed check for path and domain.
+    - >-
+      GH-1104 Fixed Cookie extensions not getting stringified properly.
+      Also added type checks for Cookie extensions, path and domain.
     - GH-1098 Added unit tests for cookie.js, header.js, request-auth.js, item.js, content-info.js.
 
 3.6.6:
--- a/lib/collection/cookie.js
+++ b/lib/collection/cookie.js
@@ -157,7 +157,7 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
              * host name that is sending the Set-Cookie header. The SDK does not perform this check, but the underlying
              * client that actually sends the request could do it automatically.
              */
-            domain: _.isString(options.domain) ? options.domain : undefined,
+            domain: options.domain,
 
             /**
              * @type {String}
@@ -165,7 +165,7 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
              * @note
              * On server, the default value for the path option is the path of the URL that sent the Set-Cookie header.
              */
-            path: _.isString(options.path) ? options.path : undefined,
+            path: options.path,
 
             /**
              * A secure cookie will only be sent to the server when a request is made using SSL and the HTTPS protocol.
@@ -208,7 +208,7 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
              * Any extra parameters that are not strictly a part of the Cookie spec go here.
              * @type {Array}
              */
-            extensions: _.isArray(options.extensions) ? options.extensions : undefined
+            extensions: options.extensions
         });
     },
 
@@ -269,9 +269,9 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
             str += '; HttpOnly';
         }
 
-        if (this.extensions) {
+        if (this.extensions && _.isArray(this.extensions)) {
             this.extensions.forEach(function (ext) {
-                if (!ext.key || !ext.value) {
+                if (_.isNil(ext.key) || _.isNil(ext.value)) {
                     return;
                 }
                 str += '; ' + ext.key + '=' + ext.value;
--- a/lib/collection/cookie.js
+++ b/lib/collection/cookie.js
@@ -157,7 +157,7 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
              * host name that is sending the Set-Cookie header. The SDK does not perform this check, but the underlying
              * client that actually sends the request could do it automatically.
              */
-            domain: options.domain,
+            domain: _.isString(options.domain) ? options.domain : undefined,
 
             /**
              * @type {String}
@@ -165,7 +165,7 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
              * @note
              * On server, the default value for the path option is the path of the URL that sent the Set-Cookie header.
              */
-            path: options.path,
+            path: _.isString(options.path) ? options.path : undefined,
 
             /**
              * A secure cookie will only be sent to the server when a request is made using SSL and the HTTPS protocol.
@@ -208,7 +208,7 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
              * Any extra parameters that are not strictly a part of the Cookie spec go here.
              * @type {Array}
              */
-            extensions: options.extensions || undefined
+            extensions: _.isArray(options.extensions) ? options.extensions : undefined
         });
     },
 
@@ -271,7 +271,10 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
 
         if (this.extensions) {
             this.extensions.forEach(function (ext) {
-                str += '; ' + ext;
+                if (!ext.key || !ext.value) {
+                    return;
+                }
+                str += '; ' + ext.key + '=' + ext.value;
             });
         }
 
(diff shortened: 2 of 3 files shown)


test/unit/cookie.test.js
```diff
@@ -384,5 +384,45 @@ describe('Cookie', function () {
...
+                extensions: [{
+                    key: 'Priority',
+                    value: 'HIGH'
+                }]
+            });
+
+            expect(cookie.toString()).equals('testCookie=testCookieVal; Max-Age=1502442248; Priority=HIGH');
+        });
+
+        it('should return single Set-Cookie header string with invalid key extensions', function () {
```
coditva, line 402: Not what the test is asserting.
```suggestion
        it('should not add extension when extension key is not specified', function () {
```

test/unit/cookie.test.js
```diff
@@ -384,5 +384,45 @@ describe('Cookie', function () {
...
+                maxAge: 1502442248,
+                extensions: [{
+                    value: 'HIGH'
+                }]
+            });
+
+            expect(cookie.toString()).equals('testCookie=testCookieVal; Max-Age=1502442248');
+        });
+
+        it('should return single Set-Cookie header string with invalid value extensions', function () {
```
coditva, line 415: Not what the test is asserting.

test/unit/cookie.test.js
```diff
@@ -384,5 +384,45 @@ describe('Cookie', function () {
...
+                key: 'testCookie',
+                value: 'testCookieVal',
+                maxAge: 1502442248,
+                extensions: [{
+                    key: 'Priority'
+                }]
+            });
+
+            expect(cookie.toString()).equals('testCookie=testCookieVal; Max-Age=1502442248');
+        });
```
coditva, line 426: Also add test when invalid data type extensions are passed as:

```javascript
extensions: [{
	key: { a: 1 },
	value: function () {}
}]
```

test/unit/cookie.test.js
```diff
@@ -384,5 +384,45 @@ describe('Cookie', function () {
...
+            var cookie = new Cookie({
+                key: 'testCookie',
+                value: 'testCookieVal',
+                maxAge: 1502442248,
+                extensions: [{
+                    key: 'Priority'
+                }]
+            });
+
+            expect(cookie.toString()).equals('testCookie=testCookieVal; Max-Age=1502442248');
```
coditva, line 425: `.to.equal()`

test/unit/cookie.test.js
```diff
@@ -384,5 +384,45 @@ describe('Cookie', function () {
...
+            var cookie = new Cookie({
+                key: 'testCookie',
+                value: 'testCookieVal',
+                maxAge: 1502442248,
+                extensions: [{
+                    value: 'HIGH'
+                }]
+            });
+
+            expect(cookie.toString()).equals('testCookie=testCookieVal; Max-Age=1502442248');
```
coditva, line 412: `.to.equal`

test/unit/cookie.test.js
```diff
@@ -384,5 +384,45 @@ describe('Cookie', function () {
...
+                key: 'testCookie',
+                value: 'testCookieVal',
+                maxAge: 1502442248,
+                extensions: [{
+                    key: 'Priority',
+                    value: 'HIGH'
+                }]
+            });
+
+            expect(cookie.toString()).equals('testCookie=testCookieVal; Max-Age=1502442248; Priority=HIGH');
```
coditva, line 399: `.to.equal`

lib/collection/cookie.js
```diff
@@ -208,7 +208,7 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
              * Any extra parameters that are not strictly a part of the Cookie spec go here.
              * @type {Array}
              */
-            extensions: options.extensions || undefined
+            extensions: _.isArray(options.extensions) ? options.extensions : undefined
```
coditva, line 211: Do we have a test that checks the behaviour when `extensions` is not an `array`?

lib/collection/cookie.js
```diff
@@ -157,15 +157,15 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
...
+            domain: _.isString(options.domain) ? options.domain : undefined,
 
             /**
              * @type {String}
              *
              * @note
              * On server, the default value for the path option is the path of the URL that sent the Set-Cookie header.
              */
-            path: options.path,
+            path: _.isString(options.path) ? options.path : undefined,
```
coditva, line 168: Let's add tests to check behaviour when `path` and `domain` are not `strings`. Ideally we should be checking if they are:
1. `null`,
2. `object`,
3. `array`,
4. `number`,
5. `function`
6. empty `string` (in this case it should be set to empty string)

CHANGELOG.yaml
```diff
@@ -1,5 +1,6 @@
 master:
   chores:
+    - GH-1104 Fixed cookie extensions not getting stringified properly. Fixed check for path and domain.
```
coditva, line 3:
```suggestion
    - GH-1104 Fixed Cookie extensions not getting stringified properly. Also added type checks for Cookie extensions, path and domain.
```

test/unit/cookie.test.js
coditva: I would suggest we add tests for all invalid data types for `key` and `value`.

coditva (changes requested):
1. Need more tests.
2. Naming tests should be based on what the test is asserting.

lib/collection/cookie.js
lucifercr07: Fixed.

test/unit/cookie.test.js
lucifercr07: Fixed.

lib/collection/cookie.js
lucifercr07: Fixed, added for both `domain` and `path`

lib/collection/cookie.js
```diff
@@ -271,7 +271,10 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
 
         if (this.extensions) {
             this.extensions.forEach(function (ext) {
-                str += '; ' + ext;
+                if (!ext.key || !ext.value) {
+                    return;
+                }
+                str += '; ' + ext.key + '=' + ext.value;
```
coditva, line 277: Should we also type-check the `key` and `value` in this case?
And should we iterate over the whole array in  the constructor itself for this?

As far as I could see, `extensions` are only used in this method, and we're already iterating over it. Adding the check here itself might help us avoid the overhead of another iteration in the constructor.

But it would become non-DRY if some other method starts using extensions like this.

@codenirvana suggestions?

coditva:
@lucifercr07 just a note: it would be awesome if we can put the commit message describing the changes in the commit like "Add unit tests to type check cookie path and domain"
rather than something which might not be useful in the future like: "Addressed review comments"

It would also make it easier to write the changelog by just looking at the commit messages to find the changes.

lib/collection/cookie.js
```diff
@@ -157,15 +157,15 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
              * host name that is sending the Set-Cookie header. The SDK does not perform this check, but the underlying
              * client that actually sends the request could do it automatically.
              */
-            domain: options.domain,
+            domain: _.isString(options.domain) ? options.domain : undefined,
```
codenirvana, line 160: On second thought, let's not drop this user data here. So, let's keep `domain`, `path`, and `extensions` as it is.

lib/collection/cookie.js
codenirvana: We don't mutate user data (if set incorrectly) in the constructor function (with few exceptions) and handle them in methods like `toString` etc.

lib/collection/cookie.js
```diff
@@ -271,7 +271,10 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
 
         if (this.extensions) {
             this.extensions.forEach(function (ext) {
-                str += '; ' + ext;
+                if (!ext.key || !ext.value) {
+                    return;
```
codenirvana, line 275: This will drop the key or value set as empty string. Check with `_.isNil`

lib/collection/cookie.js
codenirvana, line 277: For `key=` and `httpOnly` value is parsed as `true` so it will be seralized incorrectly here.

codenirvana (changes requested):

lib/collection/cookie.js
lucifercr07: Incase `key` value is there but `value` is empty we won't set in string due to the `_.isNil` check. We also have unit tests for that, please let me know if I am misunderstanding something here.

lib/collection/cookie.js
```diff
@@ -269,9 +269,14 @@ _.assign(Cookie.prototype, /** @lends Cookie.prototype */ {
...
-        if (this.extensions) {
+        if (_.isArray(this.extensions)) {
             this.extensions.forEach(function (ext) {
-                str += '; ' + ext;
+                if (_.isNil(ext.key)) {
+                    return;
+                }
+
+                var value = (ext.value) ? ext.value : true;
+                str += '; ' + ext.key + '=' + value;
```
coditva, line 279: Nitpick: You can also use a template string for better readability.
```suggestion
                str += `; ${ext.key}=${value}`;
```

Just a personal preference. Feel free to ignore. 🙃

PR #1105, opened 2020-09-11
Fix `toJSON()` of RequestBody

vikiCoder:
- Override `toJSON()` in FormParam to not return value for param with `type: 'file'`
- Override `toJSON()` in RequestBody to not return content for file body type

### DIFF
--- a/lib/collection/form-param.js
+++ b/lib/collection/form-param.js
@@ -1,5 +1,6 @@
 var _ = require('../util').lodash,
     Property = require('./property').Property,
+    PropertyBase = require('./property-base').PropertyBase,
 
     FormParam;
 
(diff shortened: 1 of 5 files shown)


lib/collection/form-param.js
```diff
@@ -44,6 +44,22 @@ _.assign(FormParam.prototype, /** @lends FormParam.prototype */ {
...
         return this.value; // can be multiple types, so just return whatever we have instead of being too clever
+    },
+
+    /**
+     * Convert the form-param to JSON compatible plain object.
+     *
+     * @returns {Object}
+     */
+    toJSON: function () {
+        var obj = Property.prototype.toJSON.apply(this);
```
codenirvana, line 55: Use `PropertyBase.toJSON` instead.

codenirvana (changes requested):

lib/collection/form-param.js
vikiCoder: Done

codenirvana (approved):

Issue #1127, opened 2020-12-13
"missing field `name` at ... " error when transpiling Collection that already includes [info] object


WilliamBerryiii:
https://github.com/postmanlabs/postman-collection/blob/6f7cf5e00bb4f7242e54726e6835550f277b7828/lib/collection/collection.js#L193

I have a postman collection which already includes an [info] object. The above function will nuke the info object as it converts the collection to JSON. I believe (naively) that wrapping the call with a 'hasOwnProperty' should avoid the issue; but I'm not confident of forward/backward compatibility against the postman schema changes over time.

Will make a PR, but I'm also conscious that the change may need more robust error handling for cases I'm not familiar with.

codenirvana:
@WilliamBerryiii Sorry for the delayed response and thank you for your contribution.
I've updated the failing CI tests in your PR.

Can you give more information about your use-case? Are you trying to add/preserve your custom props in the `info` object?

WilliamBerryiii:
I was attempting to do so (mainly the "name" field in the info element) ... but I elected to take another approach and just defined a collection.name value (root) which does propagate correctly. Maybe we just nuke my change unless you see value in it.

PR #1128, opened 2020-12-13
Add check for existing info object before migrating data from root level

WilliamBerryiii:
This closes issue #1127 by wrapping the root data migration to the 'info' object, with a hasOwnProperty('info') check on the root.


WilliamBerryiii:
The build appears to be blocked on an "invalid deflated response" for one of the Linux builds. Please let me know if this is a transient failure or was legit caused by this change. 

Also, happy to add tests to the PR, but per issue comment, I'm unsure of forward/backward schema compatibility issues that my be at play.

codenirvana:
Closing as per https://github.com/postmanlabs/postman-collection/issues/1127#issuecomment-755029620

PR #1171, opened 2021-04-08
Fixed mime detection when `Content-Type` is missing

dev-kanishk:
Issue link: https://github.com/postmanlabs/postman-app-support/issues/8377

**Postman-collection provides the wrong extension .mpga for the response type .mp3.** 

**RCA**
Sometimes the response of an API does not have a content type in the header, In such cases postman-collection figures out the suitable extension for the response by using other npm modules "file-type", "mime-types" and "mime-format".

In case of response type mp3 without any content type in the header. The file-type module is used to figure out the mime (Multipurpose Internet Mail Extensions) of the response and further mime-type module is used to figure out the extension for mime.
Here is the problem the mime of the mp3 file is audio/mpeg to which the mime-type module provides a .mpeg extension.

**Steps to reproduce:**

Run a local server using the following node-js snippet:

```
var http = require('http'),
    fileSystem = require('fs'),
    path = require('path');

http.createServer(function(request, response) {
    var filePath = path.join(__dirname, 'myfile.mp3');
    var stat = fileSystem.statSync(filePath);

    response.writeHead(200, {
        'Content-Length': stat.size
    });

    var readStream = fileSystem.createReadStream(filePath);
    // We replaced all the event handlers with a simple call to readStream.pipe()
    readStream.pipe(response);
})
.listen(2000);

```
Make a request to localhost:2000 and check the fileExtension property of content-info in response which will be mpeg.

solution: -> "file-type" module can directly provide the accurate file extension to the response.


codenirvana (changes requested):
Add tests.

test/unit/content-info.test.js
```diff
@@ -23,6 +23,23 @@ describe('contentInfo module', function () {
...
...EFAubqhUWDDUIDAkSTCsrhpcDA8pDCAOR5pTYrKTUESTCscw4HzC8kTuvDBEQEINARgZpAJEOrMwcSAXj/+5Jk54wDKT3Ik2kcoltEGPJt44YOsUMiTaRySYGM5emnraSU9qZjmfIgQxFAC8FBbwgk8y3dCJcrxNnKaZegjqz7Om35aTxJT+UjLOD31NcWRK7//7rDG5cwZQzZkd/WiEibbZrAxAjgtBPwVywD5eCaDcI0eo5BmmdZY1yY0cVYnivXeXTpxCFS1xAM3jzK0ecPBxbKvIX//ikZb//+KLSWcxC4wA2m0k4FPiAATAYShYlQ4Ah4TDK4WjNyTjoUABEpgQaVD07YoMFBC6xMryKkQNil3rdUYXftSxl1FWnhmtyC02mSk4+NSRW98/LSrOjyGCIrE/lV+4yc11Op92/wc8Rta8loQnjFtF9dXhmymwuvEA6wnJsgNPTW3ANwQAzIdOTirL5fhjzyRRTAqMB13HgVXfMiLUQd6mjG6a9jDsjRZ7+l4N2w9m1MPKiodNvaV//8h70sd0sZ/IIYdZGqXekAOUqwAjCYgApiljA5GmGwwYoQxtQJHj5Af4YGyHghNjGCA/YbCpeAgMyoQa7H0478Vnn0Jg1XGdhRvPdf//uSZNsIA6Q5yJO6GsJGAwqdPMeRjlh/KU7sx4kvCuo1hioWDvJXTaHWMCafTDBRsF4QW1nLXS/TJfzyU4OUa716qrkVznO1aVv2/53x+2r6fj657RW7uGXr60QYZHbLWwGFSIThR48esylWEKdTDV7DZreU2WzXJBXYAT4xq1DqJD4JO0qytNXQWGTHOQtHlUBQmPBscDo0Ddn0jOa////oe1h4HElyBjVRYMKgLQDAwsBEAhe8wyMYwIMEzc38wpGMwQA8DCeSEAfvpnZAKoZhlUofq87UNp2OCknudZRnvcWh2Zr+9EH3rH2t72gJyHKamhaQ4MZw7s0zopXuZHX+zkru+6IzX50QWm4b18A/qsrflt9x9X8K/+U0HzQKTtuusDECCQrOVdqTj7OGgiACv3ZjpKUTwRx8P3UpGdMLakTgQ7dprdQXyaWFA1O8lLX6TpMah5LSCwZHNtyH/oqyrt3/65RZRcAHg0VXlRWDAcHiIOwYdw8MRhKG5g8UhkSK5rlExmeDIFDcwUDUwdA0eg4wjFYeCQ40p5REZRCIYQ==',
```
codenirvana, line 30: Let's move this to `test/fixtures/audio.mp3`

test/unit/content-info.test.js
```diff
@@ -23,6 +23,23 @@ describe('contentInfo module', function () {
         });
     });
 
+    // eslint-disable-next-line max-len
+    it('For mp3 files response, it should take the sniffed content-type from the response stream if content-type and content-disposition headers is not present', function () {
```
codenirvana, line 27: `Should detect mp3 response stream if content-type is not present`

codenirvana (changes requested):

test/unit/content-info.test.js
dev-kanishk: done

test/unit/content-info.test.js
dev-kanishk: done

test/unit/content-info.test.js
```diff
@@ -23,6 +24,23 @@ describe('contentInfo module', function () {
         });
     });
 
+    // eslint-disable-next-line max-len
```
codenirvana, line 27: not required.

test/unit/content-info.test.js
```diff
@@ -23,6 +24,23 @@ describe('contentInfo module', function () {
...
+    it('Should detect mp3 response stream if content-type is not present', function () {
+        // data url of mp3 file
+        // eslint-disable-next-line max-len
```
codenirvana, line 30: not required.

test/unit/content-info.test.js
```diff
@@ -23,6 +24,23 @@ describe('contentInfo module', function () {
...
+        var data = fs.readFileSync('test/fixtures/audio.mp3'),
+            // creating the buffer of the file
+            response = new Response({ stream: Buffer.from(data, 'base64') });
```
codenirvana, line 33: Isn't `data` already a  Buffer?

codenirvana (changes requested):

test/unit/content-info.test.js
```diff
@@ -23,6 +24,23 @@ describe('contentInfo module', function () {
         });
     });
 
+    // eslint-disable-next-line max-len
+    it('Should detect mp3 response stream if content-type is not present', function () {
```
codenirvana, line 28: This will not work in browser tests.

```suggestion
    // eslint-disable-next-line max-len
    (typeof window !== 'undefined' ? it.skip : it)('Should detect mp3 response stream if content-type is not present', function () {
```

test/unit/content-info.test.js
dev-kanishk: I am now using a trimmed actual mp3 file not just the Base64 encoding of an mp3 in some file with a .mp3 extension. Reason: fs.readFileSync was reading the encoding as a text.

test/unit/content-info.test.js
dev-kanishk: it is now. after applying condition for skipping the test for browser.

test/unit/content-info.test.js
dev-kanishk: done

test/unit/content-info.test.js
codenirvana: `readFileSync` returns a buffer, this will also work:

```js
var response = new Response({
    stream: fs.readFileSync('test/fixtures/audio.mp3')
});
```

PR #1183, opened 2021-05-12
Upgrade to "sanitize-html": "2.3.3". 

gone-skiing:
Addresses https://github.com/advisories/GHSA-mjxr-4v3x-q3m4

Fix expected result, sanitize now replaces # with h1.

It appears that sanitize-html 2.X brings in postcss version that is no longer compatible with node 6. I am wondering if it is still necessary to support node 6 given the EOL date of April 2019.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ node_js:
   - 10
   - 12
   - 14
-  - 15
+  - 16
 
 jobs:
   include:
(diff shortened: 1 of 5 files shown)


gone-skiing:
Does anyone know what is going on with npm ci credential issue? It looks like it is specific to node 16.

codenirvana:
@gone-skiing Thanks for the fix but with `postman-collection` v4, we've dropped `sanitize-html` dependency all together 😅

PR #1208, opened 2021-08-04
Fix incorrect descriptions for randomInt and randomPrice

coditva:
Relates to https://github.com/postmanlabs/postman-app-support/issues/10168

### DIFF
--- a/lib/superstring/dynamic-variables.js
+++ b/lib/superstring/dynamic-variables.js
@@ -99,7 +99,7 @@ var faker = require('faker/locale/en'),
         },
 
         $randomInt: {
-            description: 'A random integer between 1 and 1000',
+            description: 'A random integer between 0 and 1000',
             generator: function () {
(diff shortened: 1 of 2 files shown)


codenirvana (approved):

lib/superstring/dynamic-variables.js
```diff
@@ -312,7 +312,7 @@ var faker = require('faker/locale/en'),
             generator: faker.finance.mask
         },
         $randomPrice: {
-            description: 'A random price between 100.00 and 999.00',
+            description: 'A random price between 0.00 and 999.00',
```
neilstudd, line 315: @coditva The correct value here is actually `between 0.00 and 1000.00` - both ends of the range were incorrect

coditva: @neilstudd Thanks for pointing out! I missed this. 🙈 

Fix in #1209

PR #1209, opened 2021-08-05
Fix incorrect description for `randomPrice` upper limit

coditva:
Ref https://github.com/postmanlabs/postman-collection/pull/1208#discussion_r682686137

### DIFF
--- a/lib/superstring/dynamic-variables.js
+++ b/lib/superstring/dynamic-variables.js
@@ -312,7 +312,7 @@ var faker = require('faker/locale/en'),
             generator: faker.finance.mask
         },
         $randomPrice: {
-            description: 'A random price between 0.00 and 999.99',
+            description: 'A random price between 0.00 and 1000.00',
             generator: faker.finance.amount
         },
         $randomTransactionType: {
--- a/lib/superstring/dynamic-variables.js
+++ b/lib/superstring/dynamic-variables.js
@@ -312,7 +312,7 @@ var faker = require('faker/locale/en'),
             generator: faker.finance.mask
         },
         $randomPrice: {
-            description: 'A random price between 0.00 and 999.00',
+            description: 'A random price between 0.00 and 999.99',
             generator: faker.finance.amount
         },
         $randomTransactionType: {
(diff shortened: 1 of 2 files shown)


lib/superstring/dynamic-variables.js
```diff
@@ -312,7 +312,7 @@ var faker = require('faker/locale/en'),
             generator: faker.finance.mask
         },
         $randomPrice: {
-            description: 'A random price between 0.00 and 999.00',
+            description: 'A random price between 0.00 and 999.99',
```
neilstudd, line 315: I promise this needs to be `1000.00`, not `999.99` 😃 

Demo from Tests tab

```
prices = []
for (i = 1; i < 100000; i++) {
    prices.push(pm.variables.replaceIn('{{$randomPrice}}'));
}
prices.sort((function(a, b){return a-b}));
console.log("Prices range from " + prices[0] + " to " + prices[prices.length-1]);
```

Results in console message `Prices range from 0.00 to 1000.00`

neilstudd (changes requested):
Made one small ($0.01) change request :)

lib/superstring/dynamic-variables.js
coditva: @neilstudd I tried this but could only find the max price to be `999.99` 🤔 
```javascript
const dynamicVars = require('./lib/superstring/dynamic-variables');

let max = 0.00;

for (let i = 0, len = 10000000; i < len; i++) {
    const price = dynamicVars.$randomPrice.generator();

    // console.log(price);

    if (max < price) {
        max = price;
    }
}

console.log(max);
```

Let me try the script you shared as well.

neilstudd: Alternate targeted example:

```
for (i = 1; i < 100000; i++) {
    if ( pm.variables.replaceIn('{{$randomPrice}}') == "1000.00") {
        console.log("found 1000");
    }
}
```

coditva: @neilstudd Aah, makes sense. 🤦🏼
Let me update this.

neilstudd (approved):

codenirvana (approved):

PR #1213, opened 2021-08-09
Added Property.findSubstitutions to get list of variables used in a property

shamasis:
```
    /**
     * This function recursively traverses a variable and detects all instances of variable replacements
     * within the string of the object
     *
     * @param {*} obj Any JS variable within which we are trying to discover {{variables}}
     * @returns {Array:String}
     *
     * @example
     * Proprty.findSubstitutions({request: 'https://{{host}}/{{path1}}-action/'});
     * -> ['host', 'path1']
     *
     * @todo
     * - tonne of scope for performance optimisations
     * - accept a reference variable scope so that substitutions can be applied to find recursive
     *   replacements (e.g. {{hello-{{hi}}-var}})
     */
```

The use-case for now is to use it in places such as UI warnings and flow runner to find what variables were used/defined in the request definition.


coditva (commented):
Looks amazing. 🚀 

- [x] The checks seem to be failing due to lower coverage but the diff hit is 100%. Need to take a look here. 🤔
- [x] Need to update change-log before we merge.

lib/collection/property.js
```diff
@@ -259,6 +276,64 @@ _.assign(Property, /** @lends Property */ {
...
+        if (_.isNil(obj) || (obj[ITERATION_LOCKED] === true)) {
+            return substitutionKeys;
+        }
+
+        // we recurse for arrays, then we recurse for objects and finally when we land on a string,
+        // we extract substitutions using regex
+        if (_.isArray(obj)) {
+            obj[ITERATION_LOCKED] = true;
+            for (let i = 0, ii = obj.length; i < ii; i += 1) {
+                substitutionKeys = substitutionKeys.concat(Property.findSubstitutions(obj[i]));
```
codenirvana, line 310: micro-optimization: array `spread` performs better than `concat`.

codenirvana (approved):
LGTM.

Maybe we can use the `Set` approach for traversing the cyclic object to avoid mutating the input value.

codenirvana (approved):

PR #1226, opened 2021-11-29
Introduce password data type

shamasis:
Password type is essentially a string type. This is also a precursor to allow external hooks into type handling.

+ added new password type
+ reordered code in type def for cleanliness
+ tests
+ no breaking change

### DIFF
--- a/lib/collection/variable.js
+++ b/lib/collection/variable.js
@@ -258,6 +258,28 @@ _.assign(Variable, /** @lends Variable */ {
         number: Number,
 
         /**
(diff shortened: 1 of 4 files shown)


lib/collection/variable.js
```diff
@@ -257,6 +257,28 @@ _.assign(Variable, /** @lends Variable */ {
          */
         number: Number,
 
+        /**
```
shamasis, line 260: This block is not a change, but merely a code reorder for clean addition of new type

PR #1241, opened 2022-01-17
chore: use @faker-js/faker instead of faker

chentsulin:
The original https://github.com/marak/Faker.js/ repo has been force pushed with an endgame commit, so we should switch to https://github.com/faker-js/faker which is maintained by the community.

close #1235


chentsulin:
Hi @codenirvana, could you get this merged and released? It's unsafe to use https://www.npmjs.com/package/faker, because its v6.6.6 release.

package.json
```diff
@@ -32,7 +32,7 @@
     "test-unit": "nyc --nycrc-path=.nycrc.js node npm/test-unit.js"
   },
   "dependencies": {
-    "faker": "5.5.3",
+    "@faker-js/faker": "^5.5.3",
```
codenirvana, line 35: Can you please pin the version and update the package-lock as well?

```suggestion
    "@faker-js/faker": "5.5.3",
```

codenirvana (changes requested):

package.json
chentsulin: Sure.

chentsulin: Done.

chentsulin:
@codenirvana Is there any chance of releasing this in a day or two? Thanks for your support! 😄

codenirvana:
@chentsulin Since the base branch was not `develop` a lot of unrelated commits showed up from `main`.
So, I directly made this change here: 2bd258d.

This will be part of the next release, I won't release this package with just this dependency bump because of the other dependent Postman packages.

chentsulin:
Ok, it makes sense. Thanks!

Issue #1267, opened 2022-04-30
URL path variables are not replaced if followed by a dot


shubhbhargav:
Reference ticket in Postman App support: https://github.com/postmanlabs/postman-app-support/issues/8302#issue-597004750

PR #1268, opened 2022-04-30
Fixed URL path extraction with media type extension

shubhbhargav:
ref:

- https://github.com/postmanlabs/postman-app-support/issues/8302
- https://github.com/postmanlabs/postman-collection/issues/1267

### DIFF
--- a/lib/collection/url.js
+++ b/lib/collection/url.js
@@ -12,8 +12,8 @@ var _ = require('../util').lodash,
     PROTOCOL_HTTP = 'http',
     HTTPS_PORT = '443',
     HTTP_PORT = '80',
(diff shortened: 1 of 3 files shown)


lib/collection/url.js
```diff
@@ -13,6 +13,7 @@ var _ = require('../util').lodash,
     HTTPS_PORT = '443',
     HTTP_PORT = '80',
     PATH_SEPARATOR = '/',
+    PATH_VARIABLE_SEPARATOR = '.',
```
coditva, line 16: non-blocker: Not sure about the variable name here. `.` isn't separating the path variables. It's just a character that shouldn't be part of the path key.

coditva (approved):
Looks good! :shipit:

codenirvana (approved):

Issue #602, opened 2018-03-12
TypeScript type definitions


kbuzby:
Hello!
I've written TypeScript type definitions and am submitting them to DefinitelyTyped: DefinitelyTyped/DefinitelyTyped#24227

I just wanted to notify here in case you would like to review and there is anything that should be updated in them, or if you would like to try and [bundle the types](http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) with your npm package (since that is the preferred method) then I would be happy to help coordinate a pull request or anything else you need to make that happen.

shamasis:
This is amazing. I will dig deep. @kamalaknn and @harryi3t will also look into it.

shamasis:
@kbuzby - is there a way you suggest where we can generate these declarations from jsdoc? We use jsdoc to generate documentation (that has type in them.) We could use something like `ted-jsdoc` to generate the types.

- https://www.npmjs.com/package/tsd-jsdoc
- https://github.com/brettle/jsdoc-to-typescript-declaration

We could put in the time to coordinate and have this in our repo and would love to have you onboard as a contributor.

kbuzby:
@shamasis I'd be more than happy to work on this - the first time I didn't really think about generating the type definitions from the documentation because I had just used the built in DefinitelyTyped type generator ([dts-gen](https://github.com/Microsoft/dts-gen)) which looks at the code to generate a starting point.

There might still be some need for manual review of the output (at least at first until the docs 100% align) to make sure the types generated match how it works and how it's _meant_ to be used.

I'll look into the two libraries you linked and see if there's anything else as well over the weekend.

Not sure about using the schema as well - it would give property information about the classes, but nothing about the method signatures (both instance and static). Maybe it could still be used in the verification process.

myahl:
Has anything come of this?

rafek1241:
@kbuzby @shamasis 
Please, **don't use types from DefinitelyTyped for now**. Why? Because it's not the same types as they currently are in the postman-collection library. I used `Collection` from this library and when I switched to `@types/postman-collection` the constructor of the `Collection` class broke. I passed the JSON object to parse into Collection type

shamasis:
Ah. We have been doing some work where we needed to do the jsDoc and types better ... that's for the "autocomplete" feature in Postman's script editors.

Perhaps those can get formally into the repo and actually add some value here?!

@codenirvana thoughts?

XVincentX:
Hey @shamasis,

[...]

PR #1271, opened 2022-05-25
add types to package.json

DetachHead:
the types won't be picked up by typescript/npm without this, so users will end up installing the outdated `@types/postman-collection` package instead:

image

artist's rendition of what the package will look like on npm after this PR is merged:

image


DetachHead:
nvm, looks like this is intentional https://github.com/postmanlabs/postman-collection/issues/602#issuecomment-740129552

PR #1294, opened 2022-11-08
Update faker version to support HTTPS urls

abelanger1:
Version 7.2 of @faker-js/faker began forcing urls to be set as https.


abelanger1:
@codenirvana Is there a reason the Postman team isn't using v7.2+ of @faker-js/faker ?

PR #1326, opened 2023-07-20
Allow passing of content in request with file type body

dev-sharma-08:
This is being done to accept buffer data in file body.

### DIFF
--- a/lib/collection/request-body.js
+++ b/lib/collection/request-body.js
@@ -214,9 +214,6 @@ _.assign(RequestBody.prototype, /** @lends RequestBody.prototype */ {
     toJSON () {
         var obj = PropertyBase.toJSON(this);
 
-        // make sure that file content is removed because it is non-serializable ReadStream
-        _.unset(obj, 'file.content');
-
         return obj;
     }
 });
--- a/lib/collection/request-body.js
+++ b/lib/collection/request-body.js
@@ -204,17 +204,6 @@ _.assign(RequestBody.prototype, /** @lends RequestBody.prototype */ {
         }
 
         return _.isEmpty(data); // catch all for remaining data modes
-    },
-
-    /**
-     * Convert the request body to JSON compatible plain object
-     *
-     * @returns {Object}
-     */
-    toJSON () {
-        var obj = PropertyBase.toJSON(this);
-
-        return obj;
     }
 });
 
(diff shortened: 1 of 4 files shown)


Sanjay-shankar (approved):
Good to go

coditva (changes requested):
@dev-sharma-08 Can we add tests for cases where non-serialisable data is passed?

test/unit/request-body.test.js
```diff
@@ -525,14 +526,24 @@ describe('RequestBody', function () {
...
+        it('should have content in file body', function () {
             var definition = {
                     mode: 'file',
-                    file: { src: 'fileSrc', content: 'this should be removed' }
+                    file: { src: 'fileSrc', content: 'this should not be removed' }
                 },
                 rBody = new RequestBody(definition);
 
-            expect(rBody.toJSON().file).to.not.have.property('content');
+            expect(rBody.toJSON().file).to.have.property('content');
```
coditva, line 536: Shouldn't we also check what is the value of `content`?

test/unit/request-body.test.js
```diff
@@ -525,14 +526,24 @@ describe('RequestBody', function () {
...
+        });
+
+        it('should correctly handle request body with non serializable data', function () {
+            var definition = {
+                    mode: 'file',
+                    file: { content: new Stream.Readable() }
+                },
+                rBody = new RequestBody(definition);
+
+            expect(rBody.toJSON()).to.eql(definition);
```
coditva, line 546: We should assert on the value of `content`.

coditva, line 546: What would the result of serialisation of a `ReadStream` be? Wouldn't it be unusable? Do we want to keep this? @codenirvana

dev-sharma-08: `toJSON()` will have no effect on the `readStream`. It will remain usable, I have checked this.
Regarding wether we want to keep it @codenirvana can share his opinion.

lib/collection/request-body.js
```diff
@@ -214,9 +214,6 @@ _.assign(RequestBody.prototype, /** @lends RequestBody.prototype */ {
     toJSON () {
```
coditva, line 214: Since we're not overriding anything in the `toJSON` function, can we remove the function override itself?

PR #1329, opened 2023-08-02
Retain string file content while parsing formdata and file bodies

codenirvana:
Closes #1326

### DIFF
--- a/lib/collection/request-body.js
+++ b/lib/collection/request-body.js
@@ -214,8 +214,10 @@ _.assign(RequestBody.prototype, /** @lends RequestBody.prototype */ {
     toJSON () {
         var obj = PropertyBase.toJSON(this);
 
-        // make sure that file content is removed because it is non-serializable ReadStream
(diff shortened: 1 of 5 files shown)


lib/collection/request-body.js
```diff
@@ -214,8 +214,10 @@ _.assign(RequestBody.prototype, /** @lends RequestBody.prototype */ {
     toJSON () {
         var obj = PropertyBase.toJSON(this);
 
-        // make sure that file content is removed because it is non-serializable ReadStream
-        _.unset(obj, 'file.content');
+        // remove value from file param if it's empty or non-string (can be non-serializable ReadStream)
+        if (obj.file && obj.file.content && typeof obj.file.content !== 'string') {
```
dev-sharma-08, line 218: Do we not want to support for buffer data itself or readstream ?

codenirvana: Any non-string value will be dropped.

PR #1333, opened 2023-09-21
Add secret variable type

dev-sharma-08:


lib/collection/mutation-tracker.js
```diff
@@ -21,7 +21,7 @@ var _ = require('../util').lodash,
      * @returns {Boolean}
      */
     isPrimitiveMutation = function (mutation) {
-        return mutation && mutation.length <= 2;
+        return mutation && mutation.length <= 3;
```
vedkribhu, line 24: why?

lib/collection/variable.js
```diff
@@ -341,6 +346,27 @@ _.assign(Variable, /** @lends Variable */ {
             }
         },
 
+        /**
+         * A "secret" type value stores data as postman secret variables
+         */
+        secret: {
```
vedkribhu, line 352: Can we remove all changes from this pr except this. Essentially decoupling the type change and secret change

vedkribhu (changes requested):

PR #1338, opened 2023-10-10
feat(property-base): add support to get path of current item

vedkribhu:


lib/collection/property-base.js
```diff
@@ -169,6 +169,28 @@ _.assign(PropertyBase.prototype, /** @lends PropertyBase.prototype */ {
         return this && this.__parent && (this.__parent.__parent || this.__parent) || undefined;
     },
 
+    /**
+     * Returns the path of the item
+     *
+     * @returns {Array<string>}
+     */
+    getPath () {
```
coditva, line 177: Why add this to `PropertyBase`? Do we want to add this to every type?

lib/collection/property-base.js
```diff
@@ -169,6 +169,28 @@ _.assign(PropertyBase.prototype, /** @lends PropertyBase.prototype */ {
...
+        const path = [];
+        let item = this;
+
+        while (item) {
+            if (item.name) {
+                path.push(item.name);
```
coditva, line 183: What if parents don't have a `.name`?

coditva: A case like:
```js
{
  name: 'collection',
  item: {
    item: {
      name: 'request'
    }
  }
}
```

lib/collection/property-base.js
```diff
@@ -169,6 +169,28 @@ _.assign(PropertyBase.prototype, /** @lends PropertyBase.prototype */ {
...
+            }
+
+            item = item.__parent;
```
coditva, line 186: Why access private property like this?

lib/collection/property-base.js
```diff
@@ -169,6 +169,28 @@ _.assign(PropertyBase.prototype, /** @lends PropertyBase.prototype */ {
...
+     * Returns the path of the item
+     *
+     * @returns {Array<string>}
+     */
+    getPath () {
+        const path = [];
+        let item = this;
+
+        while (item) {
+            if (item.name) {
```
coditva, line 182: If name is empty string?

coditva, line 182: Let's add a comment why this `if` condition is required.

coditva (changes requested):

lib/collection/item.js
```diff
@@ -329,7 +329,28 @@ _.assign(Item.prototype, /** @lends Item.prototype */ {
...
+     * Returns the path of the item
+     *
+     * @returns {Array<string>}
+     */
+    getPath: function () {
+        const path = [];
+        let item = this;
+
+        while (item) {
+            if (typeof item.name === 'string') {
```
coditva, line 344: Let's check why the `if` is needed here.

lib/collection/item.js
```diff
@@ -329,7 +329,28 @@ _.assign(Item.prototype, /** @lends Item.prototype */ {
...
+                path.push(item.name);
+            }
+
+            item = item.__parent;
+        }
+
+        return path.reverse();
     }
+
```
coditva, line 353: Extra line?

lib/collection/item.js
vedkribhu: Added comment

lib/collection/property-base.js
vedkribhu: Handled

lib/collection/property-base.js
vedkribhu: Because .parent is returning grandparent, and skipping elements.

lib/collection/property-base.js
vedkribhu: Adding empty string for such parents now

lib/collection/property-base.js
vedkribhu: Changed now

vedkribhu:
Reopening this here: https://github.com/postmanlabs/postman-collection/pull/1339 as @coditva requested changes and he is on leave currently

PR #1339, opened 2023-10-16
Added `getPath` method on `Item` class

vedkribhu:


lib/collection/item.js
```diff
@@ -329,6 +329,26 @@ _.assign(Item.prototype, /** @lends Item.prototype */ {
...
+
+    /**
+     * Returns the path of the item
+     *
+     * @returns {Array<string>}
+     */
+    getPath: function () {
+        const path = [];
+
+        if (typeof this.name === 'string') {
```
codenirvana, line 342: This is not needed.

We should add an empty-string name if it is not present.

test/unit/item.test.js
```diff
@@ -193,6 +193,40 @@ describe('Item', function () {
...
+                r1 = collection.items.members[2];
+
+            expect(r1.getPath()).to.deep.equal(['multi-level-folders-v2', 'R1']);
+        });
+
+        it('should return correct path for 3 level nested item', function () {
+            const collection = new sdk.Collection(fixtures.nestedCollectionV2),
+                f2 = collection.items.members[1],
+                f3 = f2.items.members[0],
+                r1 = f3.items.members[0];
```
codenirvana, line 216: Unused variables.

codenirvana (changes requested):

test/unit/item.test.js
vedkribhu: They aren't unused, next declaration is using previous.  Instead if I access request directly from collection it would be one long ugly chained access.

codenirvana: Check if `oneDeep` helps here?

test/unit/item.test.js
```diff
@@ -193,6 +193,54 @@ describe('Item', function () {
...
+
+            expect(req.getPath()).to.deep.equal([
+                'multi-level-folders-v2',
+                'F2',
+                'F2.F3',
+                'F2.F3.R1'
+            ]);
+        });
+
+        it('should return path as empty array for item without name', function () {
```
coditva, line 230: This is not an empty array

Littlewarrior3 (approved):

coditva (approved):

appurva21 (approved):

codenirvana (approved):

PR #1352, opened 2024-02-12
[APITEST-766] Added Packages to Collection Schema

aditya-baradwaj:
This PR adds the concept of packages to the collection schema and the postman-collection entity.
In the events object, we have a Listen (String) and Script Object to encapsulate all the data related to a script.
Inside the Script Object, we have the following -
- type
- exec
- id

We are adding a fourth entity - **Packages**

Packages is an array of objects, each of which will contain the name of the package and the ID of the package. This enables us to correctly resolve the packages needed for the execution of the particular parent event.
The events object in the collection will now look something like this -
`{
...,
"events": [
            {
                "listen": "prerequest",
                "script": {
                    "exec": [
                        "console.log('x');"
                    ],
                    "type": "text/javascript",
                    "packages": [
                      {
                        id: "<packageid-1>"
                      },
                      {
                        id: "<packageid-2>"
                      }
                    ],
                    "id": "51f1d8c7-6006-43e8-90f6-846cc8144d7f"
                }
            }
        ],
...
}`

Implementation documents -
- Runtime/Sandbox Changes - https://postmanlabs.atlassian.net/wiki/x/DAB3HAE
- Resolver changes (runtime consumer side) - https://postmanlabs.atlassian.net/wiki/x/BgELHwE

### DIFF
--- a/lib/collection/script.js
+++ b/lib/collection/script.js
@@ -62,6 +62,12 @@ _.assign(Script.prototype, /** @lends Script.prototype */ {
          * @type {string}
          */
         this.type = options.type || 'text/javascript';
+
+        /**
+         * @arguments {Script.prototype}
+         * @type {Array<Object>}
+         */
+        this.packages = options.packages || [];
         _.has(options, 'src') && (
 
             /**
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2158,15 +2158,18 @@ declare module "postman-collection" {
          * @param [options.type] - Script type
          * @param [options.src] - Script source url
          * @param [options.exec] - Script to execute
+         * @param [options.packages] - List of packages to be preloaded
          */
         update(options?: {
             type?: string;
             src?: string;
             exec?: string[] | string;
+            packages?: Object[];
         }): void;
         type: string;
         src: Url;
         exec: string[];
+        packages: Object[];
         /**
          * Check whether an object is an instance of ItemGroup.
          * @param obj - -


lib/collection/script.js
```diff
@@ -62,6 +62,12 @@ _.assign(Script.prototype, /** @lends Script.prototype */ {
          * @type {string}
          */
         this.type = options.type || 'text/javascript';
+
+        /**
+         * @arguments {Script.prototype}
+         * @type {Array<Object>}
```
coditva, line 68: This would provide more info about the interface:
```suggestion
         * @type {Array<{ id: string, name: string }>}
```

lib/collection/script.js
```diff
@@ -62,6 +62,12 @@ _.assign(Script.prototype, /** @lends Script.prototype */ {
          * @type {string}
          */
         this.type = options.type || 'text/javascript';
+
+        /**
+         * @arguments {Script.prototype}
```
coditva, line 67: Typo?
```suggestion
         * @augments {Script.prototype}
```

coditva: Do we even need this JSDoc tag? 🤔

lib/collection/script.js
```diff
@@ -62,6 +62,12 @@ _.assign(Script.prototype, /** @lends Script.prototype */ {
...
+         * @type {Array<Object>}
+         */
+        this.packages = options.packages || [];
```
coditva, line 70: Nit: newline after this.

lib/collection/script.js
```diff
@@ -62,6 +63,9 @@ _.assign(Script.prototype, /** @lends Script.prototype */ {
          * @type {string}
          */
         this.type = options.type || 'text/javascript';
+
+        this.packages = options.packages || {};
```
coditva, line 67: Let's keep the `@type` properties for packages. Needs to be done for line 52 as well.

test/unit/collection.test.js
```diff
@@ -25,7 +25,8 @@ describe('Collection', function () {
                         script: {
                             id: 'my-script-1',
                             type: 'text/javascript',
-                            exec: ['console.log("This doesn\'t matter");']
+                            exec: ['console.log("This doesn\'t matter");'],
+                            packages: {}
```
coditva, line 29: Why do we need to add the empty key everywhere? It's supposed to be optional, no?

test/unit/event-list.test.js
```diff
@@ -10,7 +10,8 @@ describe('EventList', function () {
             id: 'my-global-script-1',
             script: {
                 type: 'text/javascript',
-                exec: 'console.log("hello");'
+                exec: 'console.log("hello");',
+                packages: [{ id: 'script-package-1', name: 'package1' }]
```
coditva, line 14: Needs to be updated?

test/unit/collection.test.js
aditya-baradwaj: Yep yep, just needed to add this to the assertions after conversion since we're defaulting to an empty object, no need for this here.

lib/collection/script.js
coditva, line 67: Do we need to add a default value if this key is not present? If it's an optional property, I'm guessing... no?

aditya-baradwaj: @coditva Do we mean to make this completely optional? 
For consistency purposes, I thought it would be best to have this property in all event objects, and rather than leaving it as undefined in the cases where it isn't available, it seemed better to have it as an empty object.

We can change this in such a way that we move this under an 
if (options.packages) to make this completely optional, but it does not seem ideal IMO. Do let me know if this is not the case.

It would be completely optional from a User perspective in either way tho.

coditva: A whole lot of collections will start having empty `packages` field. I want to avoid it for the collections that don't even use it.

aditya-baradwaj: makes sense, I'll make it so that this.packages is only set if options.packages exists. Sounds good?

PR #1368, opened 2024-07-31
Marked headers list as having case insensitive index

parthverma1:


parthverma1:
Unnecessary change. Confirmed with tests

PR #1388, opened 2024-11-18
wip: add support for lazily replacing variables

Pranav2612000:


lib/collection/property.js
```diff
@@ -319,6 +339,49 @@ _.assign(Property, /** @lends Property */ {
...
+    replaceSubstitutionsInLazy: async function (obj, variables) {
+        // if there is nothing to replace, we move on
+        if (!(obj && _.isObject(obj))) {
+            return obj;
+        }
+
+        // convert the variables to a substitutor object (will not reconvert if already substitutor)
+        variables = Substitutor.box(variables, Substitutor.DEFAULT_VARS);
+
+        const promises = [];
```
Pranav2612000, line 359: @appurva21 Thoughts on supporting async updates this way? This feels weird but I didn't want to touch the `_.mergeWith` and `customizer` fns. Or do you think it's okay if we write a custom implementation for `mergeWith` which supports async replaces ?

lib/collection/variable.js
```diff
@@ -208,6 +222,7 @@ _.assign(Variable.prototype, /** @lends Variable.prototype */ {
         _.has(options, 'system') && (this.system = options.system);
         _.has(options, 'disabled') && (this.disabled = options.disabled);
         _.has(options, 'description') && (this.describe(options.description));
+        _.has(options, 'lazy') && (this.lazy = options.lazy);
```
Pranav2612000, line 225: @appurva21 Thoughts on this approach?
The advantage is that it makes it clear which variables are to be resolved lazily, and which are to be replaced immediately. The disadvantage is that `lazy` has no meaning for `replaceSubstitutions` ( non lazy ), and makes the `replaceSubstitutionsLazy` interface more complicated.

Some other approaches I thought about
- A) lazily replace all values ( when invoked through `replaceSubstitutionsLazy` ) but this increases the amount of memory needed, and slightly slows down the function.
- B) Only lazily replace `Promise` values. This is a good solution but there is no reliable method of identifying if something is a promise or not. ( Duck typing is not accurate )
- C) Only lazily replace function values. Could be a good tradeoff but has the same problems as (A), also makes the code slightly less intuitive.

lib/superstring/index.js
```diff
@@ -64,6 +64,46 @@ _.assign(SuperString.prototype, /** @lends SuperString.prototype */ {
...
+            try {
+                // eslint-disable-next-line no-await-in-loop
+                let value = await replacerFn(...match);
+
+                index += match.index;
+                this.value = this.value.slice(0, index) + value + this.value.slice(index + match[0].length);
+                index += match[0].length;
+            }
+            catch (_err) { /* empty */ }
+        }
```
Pranav2612000, line 99: custom string replace logic instead of using String.replace ( https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace ) as it does not support async replacer fn

PR #1408, opened 2025-03-19
Fixed a case where url.getPath interface was throwing typeError

### LINKED ISSUE #1407: `url.getPath(true)` interface throws type error when the there is no`/` in the url of the request
For `{{baseUrl}}` this kind of a url, the `url.getPath(true)` interface throws an error. 

Basically whenever there is no `/`  present in the url, the `path` attribute of the url object is set as undefined. And getPath throws the following error 
```
TypeError: Cannot read properties of undefined (reading 'join')\n    at PostmanUrl.getPath (/var/www-api/node_modules/postman-collection/lib/collection/url.js:288:47

```

barshan23:
Fixes #1407 

In the `url.getPath` interface when the interface was supposed to return the `unresolved` path, there was an unsafe access being made for `.join` on the `this.path` attribute. 

For urls which doesn't have `/` in them, e.g `{{baseUrl}}` this kind of a url, the `path` attribute is `undefined`. Hence `this.path.join` throws a type error.

The change fixes this unsafe access by defaulting `path` to an empty string when `this.path` is `undefined`.

### DIFF
--- a/lib/collection/url.js
+++ b/lib/collection/url.js
@@ -285,9 +285,9 @@ _.assign(Url.prototype, /** @lends Url.prototype */ {
     getPath (unresolved) {
         // for unresolved case, this is super simple as that is how raw data is stored
         if (unresolved) {
-            const path = this.path ? this.path.join(PATH_SEPARATOR) : '';
-
-            return PATH_SEPARATOR + path;
+            return PATH_SEPARATOR + (
+                this.path ? this.path.join(PATH_SEPARATOR) : E
+            );
         }
 
         var self = this,
--- a/lib/collection/url.js
+++ b/lib/collection/url.js
@@ -285,7 +285,9 @@ _.assign(Url.prototype, /** @lends Url.prototype */ {
     getPath (unresolved) {
         // for unresolved case, this is super simple as that is how raw data is stored
         if (unresolved) {
-            return PATH_SEPARATOR + this.path.join(PATH_SEPARATOR);
+            const path = this.path ? this.path.join(PATH_SEPARATOR) : '';
+
+            return PATH_SEPARATOR + path;
         }
 
         var self = this,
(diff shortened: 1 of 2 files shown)


lib/collection/url.js
```diff
@@ -285,7 +285,9 @@ _.assign(Url.prototype, /** @lends Url.prototype */ {
     getPath (unresolved) {
         // for unresolved case, this is super simple as that is how raw data is stored
         if (unresolved) {
-            return PATH_SEPARATOR + this.path.join(PATH_SEPARATOR);
+            const path = this.path ? this.path.join(PATH_SEPARATOR) : '';
```
srinivasdamam, line 288: To effectively resolve this issue, I recommend modifying `this.path` to set as an empty string rather than `undefined`. This will help prevent similar problems when handling paths in the future.

test/unit/url.test.js
```diff
@@ -1700,6 +1700,12 @@ describe('Url', function () {
                 expect(url.toString()).to.eql('https://postman-echo.com////////get/');
             });
 
+            it('getPath->unresolved should handle empty path properly', function () {
+                var url = new Url('{{host}}/');
+
+                expect(url.getPath(false)).to.be.equal('/');
```
srinivasdamam, line 1706: Shouldn't we test the `getPath(true)` since we made modifications for the true case?

lib/collection/url.js
barshan23: I'm not sure if we should be defaulting to an empty string. Since I saw in some places we're setting path as `undefined` explicitly. 
E.g [this](https://github.com/barshan23/postman-collection/blob/develop/lib/collection/url.js#L121)

Hence, I felt it would be safe enough to have this safe access check scoped to this function only and it would not interfere with other parts of the code.

@codenirvana Let me know if you feel otherwise.

srinivasdamam: Got you. I'm approving this. 
Kindly merge this after confirming the above with @codenirvana

srinivasdamam (approved):

PR #1421, opened 2026-02-04
Add `source` property to support referencing of secret type of variable.

VShingala:
## Overview

This PR adds support for `source` property in variable class to support proper secrets that only store references to where the value would be instead of the actual value.

`source` property would be used along with `secret` property of class to refer to exact source where secret is coming from.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for postman-collection 5.3.0-beta.1
+// Type definitions for postman-collection 5.2.1
 // Project: https://github.com/postmanlabs/postman-collection
 // Definitions by: PostmanLabs
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -2530,6 +2620,11 @@ declare module "postman-collection" {
              * A "object" type value stores Object data format
              */
             object = "{\"in\":\"\",\"out\":\"\"}",
+            /**
+             * A "secret" type variable indicates the value should be resolved from an external source
+             * using the `source` property.
+             */
+            secret = "{\"in\":\"\",\"out\":\"\"}",
```
appurva21: I don think this is needed

types/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for postman-collection 5.2.1
+// Type definitions for postman-collection 5.3.0-beta.1
```
appurva21: Can you regenerate types?

